PR #2, opened 2024-05-20
Bring over operator

astoycos:
Bring over the initial operator code + make sure our make targets for building + image building + local dev kind setup still work 

TODO
Setup actions for building + running integration tests


msherif1234:
can we update the pkgs and apis import path to point to this repo pkg I can commit into of ur commits if its ok ?

Issue #9, opened 2024-05-28
r.Status().Update() returns object not found from both operator and agent controllers


msherif1234:
after recent dependa-bot packages update the above function to update obj status not working for some reason need further investigation

PR #10, opened 2024-05-29
run kind targets on the new repo

msherif1234:
also fixes https://github.com/bpfman/bpfman-operator/issues/9 Thanks to @astoycos !!


.github/workflows/pull_request.yml
```diff
@@ -37,3 +37,64 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-build-codegen-
+      - name: checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      ## TODO(astoycos) Currently this is just done to ensure we have coverage
+      ## on all image builds. Ultimately we should be running the integration
+      ## tests with these locally built bytecode images.
```
astoycos, line 83: You can remove this TODO :)

test/integration/kprobe/base/bytecode.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
astoycos, line 1: Hrm so I don't think we should be duplicating all these here, let's talk more about the best way to install the example manifests when they live in their own repo, I'm thinking we could just maintain a static yaml and do something like 

`kustomize apply -k https://raw.githubusercontent.com/bpfman/bpfman/main/examples/config/base/go-xdp-counter/<FULL YAML>`

astoycos (approved):

astoycos (approved):

PR #18, opened 2024-06-03
build: BPFMAN_IMG & BPFMAN_AGENT_IMG to overwrite image

Billy99:
The bpfman-operator is setup to allow BPFMAN_IMG to overwrite the default bpfman image, and BPFMAN_AGENT_IMG to overwrite the bpfman-agent image. However, the Makefile is leveraging kustomize. kustomize can replace an image string in a yaml when it knows the k8s object layout, but these images are passed via a ConfigMap, which is opaque data. So the current implementation doesn't work. kustomize does have a ConfigMapGenrator, which can replace the contents of a ConfigMap.

The change is to:
* Change the kustomization.yaml to use a `configMapGenerator`.
* Use `sed` to replace the default images with those passed in (or just the default image if none were passed in). The `kustomize edit set image` command doesn't work.  The `sed` command is changing the file content, so rename kustomization.yaml to kustomization.yaml.env and piped the changes to kustomization.yaml.
* Add kustomization.yaml to .gitignore so changes aren't tracked by git.


Billy99:
To reproduce, run something like:

```
BPFMAN_IMG=quay.io/billy99/bpfman-test:pr-1150 make run-on-kind
```

Then use `kubectl describe pod -n bpfman bpfman-daemon-xxxxx` to view the container image that was loaded.

Billy99:
NOTE: `make run-on-kind` is currently not working. As a temp workaround, the BPFMAN_AGENT_IMG and BPFMAN_OPERATOR_IMG in Makefile need a `-amd64` appended to them. I didn't fix that because I think it needs a little more discussion on the correct fix.

Makefile
```diff
@@ -291,6 +291,10 @@ test: fmt envtest ## Run Unit tests.
 .PHONY: test-integration
 test-integration: ## Run Integration tests.
 	go clean -testcache
+	cd config/bpfman-deployment && \
+	  sed -e 's@bpfman\.image=.*@bpfman.image=$(BPFMAN_IMG)@' \
```
msherif1234, line 295: u could use `yq` to replace in place the image as well and avoid the tmp kustomize ?

msherif1234:
> NOTE: `make run-on-kind` is currently not working. As a temp workaround, the BPFMAN_AGENT_IMG and BPFMAN_OPERATOR_IMG in Makefile need a `-amd64` appended to them. I didn't fix that because I think it needs a little more discussion on the correct fix.

WDYT about the following to fix `-amd64` tag issue ?
```sh
.PHONY: load-images-kind
 load-images-kind: ## Load bpfman-agent, and bpfman-operator images into the running local kind devel cluster.
-       kind load docker-image ${BPFMAN_OPERATOR_IMG} ${BPFMAN_AGENT_IMG} --name ${KIND_CLUSTER_NAME}
+       kind load docker-image ${BPFMAN_OPERATOR_IMG}-${MULTIARCH_TARGETS} ${BPFMAN_AGENT_IMG}-${MULTIARCH_TARGETS} --name ${KIND_CLUSTER_NAME}
```

Makefile
Billy99: Yes, we discussed `yq` in the past, but decided we didn't want to add another tool. We are already doing something similar in the examples Makefile.

Billy99:
> > NOTE: `make run-on-kind` is currently not working. As a temp workaround, the BPFMAN_AGENT_IMG and BPFMAN_OPERATOR_IMG in Makefile need a `-amd64` appended to them. I didn't fix that because I think it needs a little more discussion on the correct fix.
> 
> WDYT about the following to fix `-amd64` tag issue ?
> 
> ```shell
> .PHONY: load-images-kind
>  load-images-kind: ## Load bpfman-agent, and bpfman-operator images into the running local kind devel cluster.
> -       kind load docker-image ${BPFMAN_OPERATOR_IMG} ${BPFMAN_AGENT_IMG} --name ${KIND_CLUSTER_NAME}
> +       kind load docker-image ${BPFMAN_OPERATOR_IMG}-${MULTIARCH_TARGETS} ${BPFMAN_AGENT_IMG}-${MULTIARCH_TARGETS} --name ${KIND_CLUSTER_NAME}
> ```

That's probably fine as long as that's the only place it is needed.

Makefile
msherif1234: just FYI `sed` require some magic to work on mac os, probably something like
```bash
ifeq (,$(shell which gsed 2>/dev/null))
SED ?= sed
else
SED ?= gsed
endif
```
then use `SED` everywhere

Billy99: Good to know. I don't think anyone has tried to build/run on a MAC, but should probably be added to the list.

msherif1234 (approved):

PR #19, opened 2024-06-05
fixup the operator-sdk bundle

astoycos:
- Make sure the bundle is installed in the `bpfman` namespace rather than the openshift-bpfman namespace.

- Update the SCC to be a bit more tight specifically making it run with

runAsUser:
  type: MustRunAsNonRoot

- Fixup some documentation

- Make the operator deploy our custom bpfman-restricted SCC if running in OCP


controllers/bpfman-operator/configmap.go
```diff
@@ -209,8 +229,28 @@ func bpfmanConfigPredicate() predicate.Funcs {
 	}
 }
 
+// LoadRestrictedSecurityContext loads the bpfman-restricted SCC from disk which
+// users can bind to in order to utilize bpfman in an unprivileged way.
+func LoadRestrictedSecurityContext(path string) *osv1.SecurityContextConstraints {
+	// Load static SCC yaml from disk
+	file, err := os.Open(path)
```
msherif1234, line 236: why u decided to load static file instead of setting scc obj using APIs ?

astoycos: It's just how we've been doing it for all of the objects deployed by the bpfman-operator (i.e the `bpfman-daemonset`, `csiDriver`, and now `SecurityContextConstraint`) and  IMO it usually ends up being a bit less code

astoycos:
Hold until I can manually test on OCP, cluster bot is fighting me ATM

controllers/bpfman-operator/configmap.go
astoycos: Additionally they all really didn't need to be configured at runtime at all (except for the bpfman ds) so I figured having them static on disk made sense

astoycos:
Tested on OCP 4.15 👍

msherif1234:
is there a need to write specific section in the readme for deploying on OCP cluster ?

astoycos:
It's already there right? https://github.com/bpfman/bpfman-operator#deploy-to-openshift-cluster

bundle/manifests/bpfman-operator.clusterserviceversion.yaml
```diff
@@ -268,22 +284,20 @@ spec:
...
-    and `fatal`, defaults to `info`\n- `bpfman.agent.log.level`: the log level for
-    the bpfman-agent currently supports `info`, `debug`, and `trace` \n\nThe bpfman
-    operator deploys eBPF programs via CRDs. The following CRDs are currently available,
-    \n\n- XdpProgram\n- TcProgram\n- TracepointProgram\n- KprobeProgram\n- UprobeProgram\n-
-    FentryProgram\n- FexitProgram\n\n## More information\n\nPlease checkout the [bpfman
-    community website](https://bpfman.io/) for more information."
+    get bpfman up and running quickly simply click 'install' to deploy the bpfman-operator
+    in the bpfman namespace via operator-hub.\n## Configuration\n\nThe `bpfman-config`
+    configmap is automatically created in the `bpfman` namespace and used to configure
+    the bpfman deployment.\n\nTo edit the config simply run\n\n```bash\nkubectl edit
```
Billy99, line 291: Is the bundle more OCP specific? If so, should `kubectl` be replaced with `oc`?

astoycos: So technically the operatorhub is for everyone, and community operators prod is for OCP I think we can leave it as kubectl here and add a carry commit to  https://github.com/redhat-openshift-ecosystem/community-operators-prod

cmd/bpfman-operator/main.go
```diff
@@ -46,9 +49,33 @@ var (
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(bpfmaniov1alpha1.Install(scheme))
+	utilruntime.Must(osv1.Install(scheme))
 	//+kubebuilder:scaffold:scheme
 }
 
+func getPlatformInfo(client discovery.DiscoveryInterface, cfg *rest.Config) (bool, error) {
```
Billy99, line 56: nit: This function should be renamed or add a comment. Returns true if Openshift, false otherwise.

cmd/bpfman-operator/main.go
```diff
@@ -129,10 +156,26 @@ func main() {
...
+	dc, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	if err != nil {
+		setupLog.Error(err, "can't instantiate discovery client")
+		os.Exit(1)
+	}
+
+	isOpenshift, err := getPlatformInfo(dc, mgr.GetConfig())
+	if err != nil {
+		setupLog.Error(err, "unable to determine platform")
+		os.Exit(1)
```
Billy99, line 169: Is it likely this could fail? I guess call to KubeAPI server fails. Wondering of we should bail on a failure or not?

controllers/bpfman-operator/configmap.go
```diff
@@ -43,12 +44,15 @@ import (
...
 // +kubebuilder:rbac:groups=storage.k8s.io,resources=csidrivers,verbs=get;list;watch;create;delete
+// +kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch;create;delete
 // +kubebuilder:rbac:groups=bpfman.io,resources=configmaps/finalizers,verbs=update
 
 type BpfmanConfigReconciler struct {
 	ReconcilerCommon
 	BpfmanStandardDeployment string
 	CsiDriverDeployment      string
+	RestrictedSCC            string
+	IsOpenshift              bool
```
Billy99, line 55: This struct is created in [configmap_test.go](https://github.com/bpfman/bpfman-operator/blob/main/controllers/bpfman-operator/configmap_test.go#L87), these new variables should be initialized there?

astoycos:
Poke @Billy99

cmd/bpfman-operator/main.go
astoycos: Usually if ANY setup fails in we bail here, I think the step before `mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{` would fail if for some reason the process couldn't access the API server

controllers/bpfman-operator/configmap.go
astoycos: we _should_ have a unit test there for this functionality :/ I'm going to make a follow up issue for this

astoycos: https://github.com/bpfman/bpfman-operator/issues/27

Billy99 (dismissed):
/LGTM

cmd/bpfman-operator/main.go
astoycos: I'll do this!

PR #20, opened 2024-06-05
fix kind and ocp deploy and run on targets

msherif1234:
- fix deploy and run-on-kind targets
- allow to run make commands on MacOS BSD env
- add bundle check PR workflow job

Example:
```sh
BPFMAN_AGENT_IMG=quay.io/mmahmoud/bpfman-agent:pr-1150  BPFMAN_OPERATOR_IMG=quay.io/mmahmoud/bpfman-operator:pr-1150 make build-images

BPFMAN_AGENT_IMG=quay.io/mmahmoud/bpfman-agent:pr-1150  BPFMAN_OPERATOR_IMG=quay.io/mmahmoud/bpfman-operator:pr-1150 make run-on-kind 

oc get pods -n bpfman
NAME                               READY   STATUS    RESTARTS   AGE
bpfman-daemon-hdzqg                3/3     Running   0          9m29s
bpfman-operator-55d6cb6c57-fzw4f   2/2     Running   0          9m32s
```

### DIFF
--- a/bundle/manifests/bpfman-config_v1_configmap.yaml
+++ b/bundle/manifests/bpfman-config_v1_configmap.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 data:
   bpfman.agent.healthprobe.addr: :8175
-  bpfman.agent.image: quay.io/bpfman/bpfman-agent:latest
+  bpfman.agent.image: quay.io/bpfman/bpfman-agent:latest-amd64
   bpfman.agent.log.level: info
   bpfman.agent.metric.addr: 127.0.0.1:8174
   bpfman.image: quay.io/bpfman/bpfman:latest
--- a/bundle/manifests/bpfman-config_v1_configmap.yaml
+++ b/bundle/manifests/bpfman-config_v1_configmap.yaml
@@ -12,4 +12,6 @@ data:
     millisec_delay = 10000
 kind: ConfigMap
 metadata:
+  annotations: {}
+  labels: {}
   name: bpfman-config
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -47,6 +47,19 @@ jobs:
         fail_ci_if_error: true
         verbose: true   
 
+  bundle-check:
+    runs-on: ubuntu-latest
+    name: Checking bundle up-to-date
+    steps:
+    - name: install make
+      run: sudo apt-get install make
+    - name: checkout
+      uses: actions/checkout@v3
+    - name: Verify generated bundle manifest
+      run: |
+        make bundle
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -28,3 +28,4 @@ go.work
 
 # Temporary Kustomize file
 config/bpfman-deployment/kustomization.yaml
+config/bpfman-operator-deployment/kustomization.yaml
```
Billy99, line 31: No, you don't want this. This file is showing up as changed as a result of the `$(KUSTOMIZE) edit set image`, but that file is still under source control. I was thinking about replacing the `$(KUSTOMIZE) edit set image` with the `sed` command as well for that reason, but didn't because it is working as designed. The other file was just broken.

bundle/manifests/bpfman-operator.clusterserviceversion.yaml
```diff
@@ -215,7 +215,7 @@ metadata:
       ]
     capabilities: Basic Install
     containerImage: quay.io/bpfman/bpfman-operator:v0.0.0
-    createdAt: "2024-06-04T18:15:26Z"
+    createdAt: "2024-06-05T21:00:08Z"
```
Billy99, line 218: I would just drop this change, meaningless.

bundle/manifests/bpfman-config_v1_configmap.yaml
```diff
@@ -12,4 +12,6 @@ data:
     millisec_delay = 10000
 kind: ConfigMap
 metadata:
+  annotations: {}
+  labels: {}
```
Billy99, line 16: Why were these added?

Billy99 (commented):
So does this build on the MAC? If so that is great. I never tried.

msherif1234:
> So does this build on the MAC? If so that is great. I never tried.

this solves `sed` issue I don't have mac to see know for sure if that is it or there more issues, but if anyone has mac just install `gsed` and we should be good to go

bundle/manifests/bpfman-config_v1_configmap.yaml
msherif1234: this change as well as TS change was side effect of `make bundle` will revert

.gitignore
msherif1234: ok will remove it

bundle/manifests/bpfman-operator.clusterserviceversion.yaml
msherif1234: removed

msherif1234: dropped

Billy99 (approved):
/LGTM

Thanks

PR #24, opened 2024-06-14
multiarch: Build multiarch binaries and images

Billy99:
Since bpfman-operator is moving to it's own repo, no longer build and push the bpfman-operator and bpfman-agent from the bpfman repo, but instead, build and push from the bpfman-operator repo.

Rework the Makefile such that it no longer loops through each architecture. Building all architectures is done in github actions. The Makefile still allows local cross compiling for a single architecture, just not all architectures in one command. This simplifies the Makefile and yaml files loading images don't need `-amd64` appended anywhere.


anfredette (commented):
LGTM, but @astoycos has been following this more closely than me, so I'll let him do the final approval.

.github/workflows/image-build.yaml
```diff
@@ -51,6 +47,75 @@ jobs:
...
+          else
+            echo "push_flag=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Build and push
+        id: build-push-image
+        uses: docker/build-push-action@v5
+        with:
+          platforms: linux/amd64, linux/arm64, linux/ppc64le, linux/s390x
+          push: ${{ fromJSON(steps.set-push.outputs.push_flag) }}
```
astoycos, line 93: hrm why the fromJSON() function call here?

astoycos: I would think just steps.set-push.outputs.push_flag would work

Billy99: Needed to convert from the string "true" to a boolean true/false.

.github/workflows/image-build.yaml
```diff
@@ -95,21 +162,10 @@ jobs:
           images: ${{ matrix.image.registry }}/${{ matrix.image.repository }}/${{ matrix.image.image }}
           tags: ${{ matrix.image.tags }}
 
-      - name: Build image
```
astoycos, line 98: Wait are we still actually building the bundle image anywhere?

Billy99: Yes, it's in it's own job, line 132 (https://github.com/bpfman/bpfman-operator/pull/24/files#diff-502adaf7a495a56eb3ecef9c7f4c295ce0ac343f9b0d4e2fd321910fdf8f58bdR132). I was on the fence on one job or two. There are a handful of common steps (Checkout bpfman, Install Golang, etc), but about several that are specific to bundle (Generate olm bundle on disk, Push to registry) and even more in the Build Image job that don't need to run.

astoycos: I get that it's in it's own job now and that's alright since we don't have to build it for every different platform, however in the https://github.com/bpfman/bpfman-operator/blob/fa94e4f9c4f536fc2cc92d6195c67f48fe64b442/.github/workflows/image-build.yaml#L132 job I don't see where we're building the bundle image

.github/workflows/image-build.yaml
```diff
@@ -64,19 +129,21 @@ jobs:
...
           go-version: '1.22' # yamllint disable-line rule:quoted-strings
 
-      - uses: sigstore/cosign-installer@v3.5.0
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.5.0
 
       - name: Generate olm bundle on disk
-        if: ${{ matrix.image.image == 'bpfman-operator-bundle' }}
         run: |
           make bundle
```
astoycos, line 148: This needs to be `make bundle-build`. Since the manifests in https://github.com/bpfman/bpfman-operator/tree/main/bundle should always be up to date now

.github/workflows/image-build.yaml
```diff
@@ -64,19 +129,21 @@ jobs:
...
 
-    name: Build Image (${{ matrix.image.image }})
+    name: Build Bundle Image (${{ matrix.image.image }})
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout bpfman
+        uses: actions/checkout@v4
 
-      - uses: actions/setup-go@v5
+      - name: Install Golang
```
astoycos, line 137: I don't think we'll even need this for just `make bundle-build`

Billy99: Removed

.github/workflows/image-build.yaml
Billy99: Done

PR #6, opened 2024-05-23
SDN-5007: introducing bfpapplication object

msherif1234:
Fix https://github.com/bpfman/bpfman/issues/1142

PR #31, opened 2024-06-21
Fixups for pr #6

anfredette:
Add watch on pod events back to `BpfApplicationReconciler`.  It was there but got dropped at some point.

Also, added a `strings.ToLower()` to the name `sanitize()` function.


msherif1234:
hmm not sure why the pod watch was dropped either changes LGTM

msherif1234 (approved):

PR #34, opened 2024-06-27
remove api doc gen

astoycos:
move the api doc generation bits to the core
bpfman/bpfman repo since that's where the
documentation currently lives.

Part of https://github.com/bpfman/bpfman/pull/1164


Billy99 (approved):
/lgtm

Issue #28, opened 2024-06-18
Develop a more concise naming scheme for BpfPrograms


anfredette:
The names are currently very long.  For example:
```
$ kubectl get bpfprograms.bpfman.io 
NAME                                                                                                               TYPE          STATUS         AGE
bpfapplication-sample-kprobe-try-to-wake-up-bpfman-deployment-control-plane-try-to-wake-up                         application   bpfmanLoaded   109s
bpfapplication-sample-tc-ingress-eth0-bpfman-deployment-control-plane-eth0                                         application   bpfmanLoaded   104s
bpfapplication-sample-tracepoint-syscalls-sys-enter-kill-bpfman-deployment-control-plane-syscalls-sys-enter-kill   application   bpfmanLoaded   106s
bpfapplication-sample-uprobe-malloc-bpfman-deployment-control-plane-libc-bpfman-daemon-wbktz-bpfman                application   bpfmanLoaded   97s
bpfapplication-sample-uprobe-malloc-bpfman-deployment-control-plane-libc-bpfman-daemon-wbktz-bpfman-agent          application   bpfmanLoaded   100s
bpfapplication-sample-xdp-eth0-bpfman-deployment-control-plane-eth0                                                application   bpfmanLoaded   94s
kprobe-example-bpfman-deployment-control-plane-try-to-wake-up                                                      kprobe        bpfmanLoaded   69s
tc-pass-all-nodes-bpfman-deployment-control-plane-eth0                                                             tc            bpfmanLoaded   28s
tracepoint-example-bpfman-deployment-control-plane-syscalls-sys-enter-openat                                       tracepoint    bpfmanLoaded   28s
uprobe-example-bpfman-deployment-control-plane-libc                                                                uprobe        bpfmanLoaded   58s
uprobe-example-containers-bpfman-deployment-control-plane-libc-bpfman-daemon-wbktz-bpfman                          uprobe        bpfmanLoaded   43s
uprobe-example-containers-bpfman-deployment-control-plane-libc-bpfman-daemon-wbktz-bpfman-agent                    uprobe        bpfmanLoaded   50s
xdp-pass-all-nodes-bpfman-deployment-control-plane-eth0                                                            xdp                          18s
```
[...]

PR #37, opened 2024-07-03
More concise naming scheme for BpfPrograms

anfredette:
Create a shorter unique name for `BpfProgram` objects. For *Programs, the name is `<*Program name>-<random string>`.  For `BpfApplications`, the unique name is `<BpfApplication name>-<bpf program type>-<random string>`.

Then, the identifying info for BpfProgram that used to be in the name is stored in the BpfProgram object.

Fixes: #28

### DIFF
--- a/controllers/bpfman-agent/common.go
+++ b/controllers/bpfman-agent/common.go
@@ -43,6 +43,7 @@ import (
 	"github.com/bpfman/bpfman-operator/internal"
 	gobpfman "github.com/bpfman/bpfman/clients/gobpfman/v1"
 	"github.com/go-logr/logr"
+	"github.com/google/uuid"
 	"google.golang.org/grpc"
 )
 
@@ -126,6 +127,9 @@ type bpfmanReconciler interface {
 	getNodeSelector() *metav1.LabelSelector
 	// getBpfGlobalData returns the Bpf program global variables.
 	getBpfGlobalData() map[string][]byte
+	// getAppProgramId() returns the program qualifier for the current
+	// program.  If there is no qualifier, it should return an empty string.
+	getAppProgramId() string
 }
 
 // reconcileCommon is the common reconciler loop called by each bpfman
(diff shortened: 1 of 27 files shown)


anfredette:
The new names look like the following:
```
$ kubectl get bpfprograms.bpfman.io 
NAME                                        TYPE          STATUS         AGE
bpfapplication-sample-kprobe-e9584268       application   bpfmanLoaded   7m56s
bpfapplication-sample-tc-5c64641c           application   bpfmanLoaded   7m49s
bpfapplication-sample-tracepoint-1f1a61a1   application   bpfmanLoaded   7m51s
bpfapplication-sample-uprobe-72e8858d       application   bpfmanLoaded   6m34s
bpfapplication-sample-uprobe-b4fde6c2       application   bpfmanLoaded   6m26s
bpfapplication-sample-xdp-f3ef4383          application   bpfmanLoaded   6m18s
fentry-example-cd940fa0                     fentry        bpfmanLoaded   6m45s
fexit-example-fd53742b                      fexit         bpfmanLoaded   7m45s
kprobe-example-f24ff1d3                     kprobe        bpfmanLoaded   7m36s
tc-pass-all-nodes-563ab6b1                  tc            bpfmanLoaded   7m16s
tracepoint-example-4a04a14a                 tracepoint    bpfmanLoaded   7m26s
uprobe-example-17e5ab9b                     uprobe        bpfmanLoaded   6m55s
uprobe-example-containers-9412fff0          uprobe        bpfmanLoaded   6m48s
uprobe-example-containers-d6b84e71          uprobe        bpfmanLoaded   6m51s
xdp-pass-all-nodes-c6626947                 xdp           bpfmanLoaded   7m5s```

File: controllers/bpfman-agent/common.go [hunk 1]
```diff
@@ -519,32 +528,47 @@ func (r *ReconcilerCommon) getExistingBpfPrograms(ctx context.Context,
 		return nil, err
 	}
 
-	existingBpfPrograms := map[string]bpfmaniov1alpha1.BpfProgram{}
+	existingBpfPrograms := map[bpfProgKey]bpfmaniov1alpha1.BpfProgram{}
 	for _, bpfProg := range bpfProgramList.Items {
-		existingBpfPrograms[bpfProg.GetName()] = bpfProg
+		key := bpfProgKey{
+			appProgId:   bpfProg.GetLabels()[internal.AppProgramId],
+			attachPoint: bpfProg.GetAnnotations()[internal.BpfProgramAttachPoint],
```
msherif1234, line 535: just curious why u need both progId and attachement point as key, progId should be unique right ?

anfredette: Not exactly.  I'm trying to use appProgId to uniquely identify a single program item within an ApplicationProgram.  For this, I'm currently using program type plus all or a portion of the attach point.  In some cases (Fentry, Fexit, Kprobe) it's the whole attach point.  In the case of uprobes, it's the FunctionName (I think I should also add the target), but it can actually be attached to the same point in multiple containers (each one results in a separate BpfProgram object).  In the case of tracepoint, it's the first tracepoint on on the list.  In the case of XDP, it's the first interface on the list.  In the case of TC, it's the first interface on the list plus the direction.  This kind of works because we've said that there will only ever be one program of a given type per attach point.  However, that may not hold true for everyone, and for the program types where there is a list of attach points, I wonder if we'd ever have two programs of the same type that have different lists, but the first item is the same. That probably won't happen with our applications, but I don't know if we can rule it out.  My assumption is that we'll need to change something in the future to make this 100% unique.  One Idea I have is to have the operator controller generate a unique ID for each program.  Another is to require that the user providd a unique name/id for each program in a BpfApplication.

anfredette: So, to summarize, the appProgId should uniquely identify a program item in a BpfApplication, and the following set of info uniquely identifies a BpfProgram object that was created for it:

Labels:
- internal.BpfProgramOwner
- internal.AppProgramId
- internal.K8sHostLabel

Annotation:
- internal.BpfProgramAttachPoint

I put the attach point in an annotation because we don't need to use it in a search, and it can be a longer string if needed.

msherif1234 (approved):

PR #41, opened 2024-07-09
cut v0.5.0-rc1 release

anfredette:
This is dependent on https://github.com/bpfman/bpfman/pull/1185 since the new version of the bpfman go grpc api bindings github.com/bpfman/bpfman v0.5.0-rc1 need to exist, CI on this will fail until that package exists.


anfredette:
I have that question too.  I'm thinking we should, but just haven't done it
before.

astoycos:
> why we don't create changes log for this repo like we do for `bpfman` ?

We could 100% especially since they're separated, however we need some sort of policy on release relationships, i.e do we always cut a bpfman-operator release alongside a bpfman release? Or is there allowed to be some version skew? 

Additionally you could easily copy over the release automation https://github.com/bpfman/bpfman/blob/main/.github/workflows/build.yml#L441-L474 which will take a checked in change-log and use it in a release on tag push

msherif1234 (approved):

PR #42, opened 2024-07-10
update build-release-yamls target

msherif1234:
this target needs to be done in bpfman repo instead of bpfman-operator


anfredette (approved):
LGTM

PR #62, opened 2024-07-23
Fix OOM because of watching configmap resource

### LINKED ISSUE #56: Reduce operator memory consumption (OOMKilled)
The bpfman operator seems to be "vulnerable" to OOMKills due to watching configmaps cluster-wide.
Steps to reproduce:

- install bpfman (just the operator is needed)
- run the following:

```bash
kubectl create namespace test
for i in {0..400}; do kubectl create cm test-cm-$i -n test --from-file=(put large file here) ; done
```
I used a file of ~ 500KB

Monitor pod metrics: ingress bandwidth and memory

image

We see that the created configmaps impact the operator memory, despite them being apparently unrelated to the operator, and deployed in a separate namespace. Then the pods gets OOM-killed.

I think this is due to missing configuration in the manager cache, that leads to watching cluster-wide configmaps (with informers used under the cover without restriction of scope).

This page touch-bases on the problem: https://master.sdk.operatorframework.io/docs/best-practices/designing-lean-operators/
[Here](https://github.com/bpfman/bpfman-operator/blob/main/controllers/bpfman-operator/configmap.go#L62-L63), the config controller sets up a watch on configmaps. Note that the Predicate function provided does not reduce the informers scope (it only reduce the generated reconcile requests), so despite of this predicate, all cluster configmaps are still set in cache under the hood.

My guess is that it should be OK to add this config options to the manager:

```go
  Cache: cache.Options{
    ByObject: map[client.Object]cache.ByObject{
      &corev1.ConfigMap{}: cache.ByObject{
	Fields: labels.SelectorFromSet(fields.Set{"metadata.name": internal.BpfmanConfigName}),
      },
    },
  },
```

(I haven't tested)

msherif1234:
Fix #56 reproduce steps are also doc in the issue

- before the fix
```sh
 oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=cri-containerd-499c6676bdf79e4157b694d4730d50eec01501a8c4faea1395a1fc62e2d52b73.scope,mems_allowed=0,oom_memcg=/system.slice/docker-113354b7db17a99a283c23fa84fb21b639ee29a2ec5a747c7f7c033a5e19b39f.scope/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-podfd0578d2_fe8d_4c8d_ba5f_20c00dede626.slice/cri-containerd-499c6676bdf79e4157b694d4730d50eec01501a8c4faea1395a1fc62e2d52b73.scope,task_memcg=/system.slice/docker-113354b7db17a99a283c23fa84fb21b639ee29a2ec5a747c7f7c033a5e19b39f.scope/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-burstable.slice/kubelet-kubepods-burstable-podfd0578d2_fe8d_4c8d_ba5f_20c00dede626.slice/cri-containerd-499c6676bdf79e4157b694d4730d50eec01501a8c4faea1395a1fc62e2d52b73.scope,task=bpfman-operator,pid=2304220,uid=65532
[534829.056623] Memory cgroup out of memory: Killed process 2304220 (bpfman-operator) total-vm:1398540kB, anon-rss:129324kB, file-rss:26880kB, shmem-rss:0kB, UID:65532 pgtables:440kB oom_score_adj:998

```

- After the fix
```sh
   PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                            
2272755 65532     20   0 1267076  39652  28032 S   0.3   0.1   0:02.86 bpfman-operator      
```

### DIFF
--- a/cmd/bpfman-operator/main.go
+++ b/cmd/bpfman-operator/main.go
@@ -27,6 +27,8 @@ import (
 
 	osv1 "github.com/openshift/api/security/v1"
 	"go.uber.org/zap/zapcore"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/discovery"
@@ -34,6 +36,8 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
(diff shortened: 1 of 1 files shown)


anfredette (approved):
LGTM

Issue #74, opened 2024-07-29
bpfman agent crashes when trying to delete unloaded BpfApplication program


anfredette:
This issue was seen when an xdp interface name was changed from an invalid interface to a valid interface.

PR #75, opened 2024-07-29
bpfman-agent: don't try to unload bpf program that isn't loaded

anfredette:
Fixes: #74

### DIFF
--- a/controllers/bpfman-agent/common.go
+++ b/controllers/bpfman-agent/common.go
@@ -670,13 +670,16 @@ func (r *ReconcilerCommon) unLoadAndDeleteBpfProgramsList(ctx context.Context, b
 			r.Logger.Error(err, "Failed to get bpf program ID")
 			return ctrl.Result{}, nil
 		}
-		r.Logger.Info("Calling bpfman to unload program on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
-		if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
-			if strings.Contains(err.Error(), programDoesNotExistErr) {
-				r.Logger.Info("Program not found on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
-			} else {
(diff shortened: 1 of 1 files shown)


controllers/bpfman-agent/common.go
```diff
@@ -670,13 +670,16 @@ func (r *ReconcilerCommon) unLoadAndDeleteBpfProgramsList(ctx context.Context, b
...
 		}
-		r.Logger.Info("Calling bpfman to unload program on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
-		if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
-			if strings.Contains(err.Error(), programDoesNotExistErr) {
-				r.Logger.Info("Program not found on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
-			} else {
-				r.Logger.Error(err, "Failed to unload Program")
-				return ctrl.Result{RequeueAfter: retryDurationAgent}, nil
+
+		if id != nil {
```
msherif1234, line 674: what about returning error here https://github.com/bpfman/bpfman-operator/blob/main/controllers/bpfman-agent/internal/bpfman-core.go#L213 instead ?

anfredette: I think that could be made to work, but I'm not sure why it was implemented the way it was.  The way GetID is currently implemented, it tells you if it wasn't set at all and if it was set, whether there was an error.  We don't handle all these cases in the code, but we should think about it.  I'd like to be able to call ListBpfmanPrograms() to see if it's loaded, but you have to know the program type.  That can be fixed, but it will take a little time.  This is a quick fix that works.  We can make it better in the future.

msherif1234: sure I just found it confusing to get invalid Id will error is nil that is why suggesting the above I am fine with what you have for now maybe worth a tracking issue so we won't forget about it
/LGTM

msherif1234 (approved):

PR #82, opened 2024-08-01
Add catalog index file and script to generate it

OlivierCazade:
The script generate a catalog index file using the current bundle.

The bundle must be reachable from a repository, it can not only be built localy.


anfredette:
@OlivierCazade, you'll need to sign off on your commit to pass the DCO checks.

Makefile
```diff
@@ -361,6 +361,10 @@ ifneq ($(origin CATALOG_BASE_IMG), undefined)
 FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
 endif
 
+.PHONY: catalog-update
+catalog-update:
```
msherif1234, line 365: pls add `## Generate catalog yaml for konflux` or something like that to show in 
`make help`

catalog.Dockerfile
```diff
@@ -5,9 +5,9 @@ ENTRYPOINT ["/bin/opm"]
 CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
 
 # Copy declarative config root into image at /configs and pre-populate serve cache
-ADD bpfman-operator-catalog /configs
+ADD catalog /configs
```
msherif1234, line 8: if u will use catalog as folder then pls delete `bpfman-operator-catalog` folder

msherif1234 (commented):
Thanks @OlivierCazade for helping out with this

hack/update_fbc.sh
```diff
@@ -0,0 +1,75 @@
...
+  echo "yq is required"
+  exit 1
+fi
+
+: "${CATALOG_FILE:=catalog/index.yaml}"
+
+cat <<EOF >"${CATALOG_FILE}"
+---
+defaultChannel: latest
+icon:
```
anfredette, line 41: What icon are you using?  Does it need to be embedded in the script like this, or could you upload and point to the svg file?

msherif1234: this bpfman log and it can be verified using https://base64.guru/converter/decode/image
and other operators doing the same the also embedded that way so we are following alone

OlivierCazade: I used the one from the bundle:
https://github.com/bpfman/bpfman-operator/blob/c596fe23739ce9d7950fe6cd9f48543b9ad640b5/bundle/manifests/bpfman-operator.clusterserviceversion.yaml#L407

I can include it in a separate file and make the script use it if you prefer.

msherif1234: no that is good enough Thanks

msherif1234 (approved):

hack/update_fbc.sh
anfredette: This is good.  Thanks.

PR #102, opened 2024-08-13
TCX ebpf hook support

### LINKED ISSUE #92: Operator support for TCX

msherif1234:
fixes #92 
- define few APIs for TCX program type
- add TCX operator and agent controller and ut
- add TCX support to bpfapplication controller
- add TCX intg test

depends on: https://github.com/bpfman/bpfman/pull/1222


apis/v1alpha1/tcxProgram_types.go
```diff
@@ -0,0 +1,93 @@
...
+const (
+	// TCXFirst TCX program will be the first program to execute.
+	TCXFirst TCXPositionType = "First"
+
+	// TCXLast TCX program will be the last program to execute.
+	TCXLast TCXPositionType = "Last"
+)
+
+// TcxProgramInfo defines the tc program details
+type TcxProgramInfo struct {
```
anfredette, line 65: We should also include revision.

anfredette (changes requested):
Looks good except for the couple of comments.  I think we should consider adding a priority option even if we don't support it right away.  I like the idea of using labels to define an order, but I just don't see how it would work for tcx programs included in a BpfApplication CRD.

apis/v1alpha1/tcxProgram_types.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Direction specifies the direction of traffic the tcx program should
+	// attach to for a given network device.
+	// +kubebuilder:validation:Enum=ingress;egress
+	Direction string `json:"direction"`
+
+	// Anchor defines the order of the TCX relative to other TCX attached to the same interface
+	// +kubebuilder:validation:Enum:="First";"Last"
+	//+kubebuilder:default:=Last
+	Position TCXPositionType `json:"position"`
```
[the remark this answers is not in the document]
msherif1234: that is the after/before ID/FD right , we dom't want the priority tc alike way right ?

apis/v1alpha1/tcxProgram_types.go
msherif1234: yes I can add it and TBD how to make sense of it

anfredette: I think it works like it's supposed to work.  We're just not sure how we would use it yet.

msherif1234: I don't think using pri for tcx is right API given how cilium and libbpf apis look like

anfredette: I'm for bpfman, itself, supporting an API that is similar to the ones supported by cilium and libbpf, but I don't think this API works well for a Kubernetes cluster.

config/samples/bpfman.io_v1alpha1_tcx_pass_tcprogram.yaml
```diff
@@ -0,0 +1,25 @@
...
+  bpffunctionname: pass
+  # Select all nodes
+  nodeselector: {}
+  interfaceselector:
+    primarynodeinterface: true
+  priority: 0
+  direction: ingress
+  bytecode:
+    image:
+      url: quay.io/bpfman-bytecode/tc_pass:latest
```
anfredette, line 17: You can change this to `quay.io/bpfman-bytecode/tcx_test:latest` now with `bpffunctionname: tcx_pass`

msherif1234:
intg test seems to broken because of https://github.com/bpfman/bpfman/pull/1324

anfredette:
It definitely was, but I think I backed out the changes that were
causing the issue.  I could be wrong, but I suspect that it's now failing
because https://github.com/bpfman/bpfman/pull/1222 added tcx to the app
counter example.

test/integration/tcx_test.go
```diff
@@ -0,0 +1,59 @@
...
+	require.Eventually(t, func() bool {
+		logs, err := req.Stream(ctx)
+		require.NoError(t, err)
+		defer logs.Close()
+		output := new(bytes.Buffer)
+		_, err = io.Copy(output, logs)
+		require.NoError(t, err)
+		t.Logf("counter pod log %s", output.String())
+
+		return doTcCheck(t, output)
```
anfredette, line 57: doTcxCheck?

controllers/bpfman-operator/tcx-program.go
```diff
@@ -0,0 +1,114 @@
...
+//+kubebuilder:rbac:groups=bpfman.io,resources=tcxprograms,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=bpfman.io,resources=tcxprograms/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=bpfman.io,resources=tcxprograms/finalizers,verbs=update
+
+func (r *TcxProgramReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.Logger = log.FromContext(ctx)
+
+	tcxProgram := &bpfmaniov1alpha1.TcxProgram{}
+	if err := r.Get(ctx, req.NamespacedName, tcxProgram); err != nil {
+		// list all TcProgram objects with
```
anfredette, line 68:
```suggestion
		// list all TcxProgram objects with
```

test/integration/tcx_test.go
anfredette: It seems like we need a doTcxCheck() function.  This is probably working because the tc test is also running and there are logs from that test.

msherif1234: I reused the Tc check since both look for the same string might be better to have one for tcx will do

test/integration/tcx_test.go
```diff
@@ -0,0 +1,59 @@
...
+		daemon, err := env.Cluster().Client().AppsV1().DaemonSets(tcxGoCounterUserspaceNs).Get(ctx, tcxGoCounterUserspaceDsName, metav1.GetOptions{})
+		require.NoError(t, err)
+		return daemon.Status.DesiredNumberScheduled == daemon.Status.NumberAvailable
+	},
+		// Wait 5 minutes since cosign is slow, https://github.com/bpfman/bpfman/issues/1043
+		5*time.Minute, 10*time.Second)
+
+	pods, err := env.Cluster().Client().CoreV1().Pods(tcxGoCounterUserspaceNs).List(ctx, metav1.ListOptions{LabelSelector: "name=go-tcx-counter"})
+	require.NoError(t, err)
+	gotcCounterPod := pods.Items[0]
```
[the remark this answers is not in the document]
msherif1234: sure

anfredette (changes requested):
Mostly a few nits.  I think the only real issue is the doTcxCheck() comment.  Otherwise, this is looking good.

apis/v1alpha1/tcxProgram_types.go
```diff
@@ -0,0 +1,84 @@
...
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster
+
+// TcxProgram is the Schema for the TcxProgram API
+// +kubebuilder:printcolumn:name="BpfFunctionName",type=string,JSONPath=`.spec.bpffunctionname`
+// +kubebuilder:printcolumn:name="NodeSelector",type=string,JSONPath=`.spec.nodeselector`
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[0].reason`
+// +kubebuilder:printcolumn:name="Direction",type=string,JSONPath=`.spec.direction`,priority=1
+// +kubebuilder:printcolumn:name="InterfaceSelector",type=string,JSONPath=`.spec.interfaceselector`,priority=1
+// +kubebuilder:printcolumn:name="Position",type=string,JSONPath=`.spec.position`,priority=1
```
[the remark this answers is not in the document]
msherif1234: ok I had b4 we added pri to tcx will include

msherif1234:
> Mostly a few nits. I think the only real issue is the doTcxCheck() comment. Otherwise, this is looking good.
This isn't an issue since both log the same strings but I created one for tcx check to be more specific

test/integration/tcx_test.go
anfredette: Actually, I just looked, and it doesn't.  I think it should, though.  That's probably another reason I changed the TC, TCX, and XDP logs in the app counter example.

anfredette: We can do that in a separate pr, though.

anfredette: Or not.  They're all in separate pods, so I guess it's clear.  Sorry for the stream of consciousness as I remember how all this stuff works :-).

anfredette (approved):
/LGTM

PR #146, opened 2024-09-10
Update images names in the bundle CSV when using konflux

msherif1234:


hack/update_bundle.sh
anfredette, line 1: I suggest adding "openshift" to the name of this script to make it more clear what it's used for.  For example: "update_bundle_openshif.sh".

hack/update_bundle.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+set -eu
+
+#!/usr/bin/env bash
+
+export BPFMAN_AGENT_IMAGE_PULLSPEC="quay.io/redhat-user-workloads/ocp-bpfman-tenant/bpfman-operator/bpfman-agent:c96ca94adc23ae27d5c6a2b20897a7f50767d8fe"
```
anfredette, line 6: Can you make the SHA a variable?

anfredette (commented):
This is already merged, but here are a couple of suggestions.

PR #156, opened 2024-09-11
update catalog pipeline and containerfile to fix violations

msherif1234:


ralphbean:
I suspect that none of these changes are necessary.

PR #209, opened 2024-09-19
make catalog patching more hermetic friendly

msherif1234:

PR #210, opened 2024-09-19
RPM prefetch for the catalog build

ralphbean:
This is based on https://github.com/bpfman/bpfman-operator/pull/209.


msherif1234:
we need to delete https://github.com/bpfman/bpfman-operator/blob/main/requirements-build.txt too right ?

msherif1234 (approved):

PR #324, opened 2024-10-16
Add separate workflow to run Kubernetes integration tests

anfredette:
Runs for pull requests, on demand, and periodically.

This allows us to easily run an integration test on a new version of bpfman, and also runs the integration tests periodically so that they run with the latest bfpman even if there are no changes in there are no changes in bpfman-operator.

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -1,0 +1,82 @@
+name: Kubernetes Integration Tests
+
+on:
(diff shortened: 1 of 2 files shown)


msherif1234:
we still also need to run intg test as part of the pull workflow otherwise it will be too late to catch issues u can invoke intg wf from build pr

msherif1234 (approved):

PR #327, opened 2024-10-22
OCPBUGS-42593: hardening service account automount

msherif1234:
Address security warning about `automount service account token – set to FALSE`

### DIFF
--- a/controllers/bpfman-operator/configmap.go
+++ b/controllers/bpfman-operator/configmap.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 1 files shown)


controllers/bpfman-operator/configmap.go
```diff
@@ -321,6 +322,7 @@ func LoadAndConfigureBpfmanDs(config *corev1.ConfigMap, path string) *appsv1.Dae
 	staticBpfmanDeployment.Spec.Template.ObjectMeta.Annotations["bpfman.io.bpfman.agent.metricaddr"] = bpfmanMetricAddr
 	staticBpfmanDeployment.Name = internal.BpfmanDsName
 	staticBpfmanDeployment.Namespace = config.Namespace
+	staticBpfmanDeployment.Spec.Template.Spec.AutomountServiceAccountToken = ptr.To(true)
```
anfredette, line 325: This may be a stupid question, but I'm trying to understand this change.

If we weren't automounting the service account token before, were we manually mounting it or passing it in somehow?  And, If so, should that mechanism be removed?

msherif1234: this is a valid question for the manual token creation there should have been 
`oc create token <token-name>` somewhere in the process but I couldn't find that 
so to me this looked like a bug that was waiting to happen when trying to mount secrets volume `/var/run/secrets/kubernetes.io/serviceaccount` for specific serviceAccount

anfredette (approved):
/LGTM

Issue #328, opened 2024-10-23
panic in bpfman-agent when trying to load XDP program


Billy99:
Author of [bpfman-Issue#1311](https://github.com/bpfman/bpfman/issues/1311) is trying to load an XDP program on OpenShift for the first time and has had several issues. Based on the logs (full logs attached below), bpfman-agent has a panic. Here is a snippet:

```
[...]

PR #330, opened 2024-10-23
panic in bpfman-agent due to err not being returned

Billy99:
Originally reported in [bpfman:#1311](https://github.com/bpfman/bpfman/issues/1311) and tracked in #328, `GetID()` and `reconcileBpfProgram()` both have instances where they detect errors but return nil for err, causing the calling functions to continue and eventually panic.

Resolves: #328

### DIFF
--- a/controllers/bpfman-agent/common.go
+++ b/controllers/bpfman-agent/common.go
@@ -217,7 +217,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 		id, err := bpfmanagentinternal.GetID(bpfProgram)
 		if err != nil {
 			r.Logger.Error(err, "Failed to get bpf program ID")
-			return bpfmaniov1alpha1.BpfProgCondNotLoaded, nil
+			return bpfmaniov1alpha1.BpfProgCondNotLoaded, err
 		}
 		switch shouldBeLoaded {
 		case true:
@@ -232,7 +232,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 				r.Logger.V(1).Info("bpf program is in wrong state, unloading and reloading", "reason", reasons, "bpfProgram Name", bpfProgram.Name, "bpf program ID", id)
 				if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
 					r.Logger.Error(err, "Failed to unload BPF Program")
-					return bpfmaniov1alpha1.BpfProgCondNotUnloaded, nil
+					return bpfmaniov1alpha1.BpfProgCondNotUnloaded, err
 				}
 
 				r.Logger.Info("Calling bpfman to load bpf program on Node", "bpfProgram Name", bpfProgram.Name)
@@ -251,7 +251,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 			r.Logger.Info("Calling bpfman to unload program on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
 			if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
 				r.Logger.Error(err, "Failed to unload Program")
-				return bpfmaniov1alpha1.BpfProgCondNotUnloaded, nil
+				return bpfmaniov1alpha1.BpfProgCondNotUnloaded, err
 			}
 		}
 	case false:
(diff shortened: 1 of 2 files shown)


anfredette:
@Billy99 You need to rebase this to pick up the changes in the Tcx pr so that integration test will pass.

anfredette:
> @Billy99 You need to rebase this to pick up the changes in the Tcx pr so that integration test will pass.

It looks like you just did that.

msherif1234 (commented):
I wonder if those cases the where err was ignored for the reconciler to retry again ?

Billy99:
> I wonder if those cases the where err was ignored for the reconciler to retry again ?

Good point @msherif1234 , I'm glad you brought that up. I meant to mention something similar when I pushed the PR but it failed some tests early and I got sidetracked. If that is the case, I think it should be up to the calling function to decide on the retry because they have more context and can just ignore the error. Let me know if you have any suggestions, you and @anfredette know the agent code a little better than me.

anfredette:
I'm looking at it now.  I had a similar thought/concern.  We need to make sure that we retry when appropriate.  It's also the case that a delete can fail if the program has already been deleted or was never installed, so we need to handle that too.

controllers/bpfman-agent/common.go
```diff
@@ -217,7 +217,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 		id, err := bpfmanagentinternal.GetID(bpfProgram)
 		if err != nil {
 			r.Logger.Error(err, "Failed to get bpf program ID")
-			return bpfmaniov1alpha1.BpfProgCondNotLoaded, nil
+			return bpfmaniov1alpha1.BpfProgCondNotLoaded, err
```
anfredette, line 220: If the ID didn't get set by now, how is it going to get set?  I don't think it is.  Did the users scenario hit this case?  However, we think the program is loaded becasue it's in the loadedBpfPrograms list, and the element in that list is the gobpfman.ListResponse_ListResult returned by bpfman, and that has all the info on the program, including the ID.  So, maybe we should try to pull it out of there.

controllers/bpfman-agent/common.go
```diff
@@ -232,7 +232,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 				r.Logger.V(1).Info("bpf program is in wrong state, unloading and reloading", "reason", reasons, "bpfProgram Name", bpfProgram.Name, "bpf program ID", id)
 				if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
 					r.Logger.Error(err, "Failed to unload BPF Program")
-					return bpfmaniov1alpha1.BpfProgCondNotUnloaded, nil
+					return bpfmaniov1alpha1.BpfProgCondNotUnloaded, err
```
anfredette, line 235: It seems reasonable to return an error here.  However, as the code stands, I think we will only try this one more time (unless a change to another program kicks off the reconciler again.)  If it fails the second time, we'll stop.

controllers/bpfman-agent/common.go
```diff
@@ -251,7 +251,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 			r.Logger.Info("Calling bpfman to unload program on node", "bpfProgram Name", bpfProgram.Name, "Program ID", id)
 			if err := bpfmanagentinternal.UnloadBpfmanProgram(ctx, r.BpfmanClient, *id); err != nil {
 				r.Logger.Error(err, "Failed to unload Program")
-				return bpfmaniov1alpha1.BpfProgCondNotUnloaded, nil
+				return bpfmaniov1alpha1.BpfProgCondNotUnloaded, err
```
anfredette, line 254: Same comment as for the UnloadBpfmanProgram() case above.

anfredette (changes requested):
I'm marking this as "Request changes" so it doesn't get merged quite yet, but I didn't actually request any changes -- just added a few comments/questions.

controllers/bpfman-agent/common.go
```diff
@@ -267,7 +267,7 @@ func (r *ReconcilerCommon) reconcileBpfProgram(ctx context.Context,
 			r.progId, err = bpfmanagentinternal.LoadBpfmanProgram(ctx, r.BpfmanClient, loadRequest)
 			if err != nil {
 				r.Logger.Error(err, "Failed to load Program")
-				return bpfmaniov1alpha1.BpfProgCondNotLoaded, nil
+				return bpfmaniov1alpha1.BpfProgCondNotLoaded, err
```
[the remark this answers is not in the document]
Billy99: Based on the logs, I think this is the only case that was hit by the user. It failed, r.progId is set to nil, no error returned to calling function and r.progId being nil caused a panic later. I returned err on all other cases just to be safe.

controllers/bpfman-agent/common.go
anfredette: We've talked about this being an issue, so I wanted to flag it for discussion.  Maybe we should keep trying, or at least try a few times.  We could count how many times it has failed in the CRD.  Maybe do some type of retry with a back off so we keep trying every once in a while.  If we could get a failure reason from bpfman, maybe we can determine whether it's a permanent or possibly transient failure to guide whether/when to do a retry.

anfredette: Ok.  I think these changes are all good.  I also think we should evaluate whether we should have retries when the bpfman calls fail, and how to do it.

anfredette (approved):
/LGTM

PR #338, opened 2024-11-04
interface's based hooks don't allow more than one function per intf per direction

msherif1234:
bpfman agent operator go into loop loading/unreloading the program incorrectly when more than on hook is used on the same interface for the same direction
for example this yaml will cause the issue to happen

```yaml
kind: BpfApplication
metadata:
  labels:
    app: netobserv
  name: netobserv
spec:
  bytecode:
    image:
      url: quay.io/bpfman-bytecode/netobserv:latest
  globaldata:
    enable_dns_tracking: AQ==
    enable_network_events_monitoring: AQ==
    enable_rtt: AQ==
    sampling: AQAAAA==
  nodeselector: {}
  programs:
  - tcx:
      bpffunctionname: tcx_ingress_flow_parse
      direction: ingress
      interfaceselector:
        interfaces:
        - eth0
      priority: 0
    type: TCX
  - tcx:
      bpffunctionname: tcx_egress_flow_parse
      direction: egress
      interfaceselector:
        interfaces:
        - eth0
      priority: 0
    type: TCX
  - tcx:
      bpffunctionname: tcx_ingress_pca_parse
      direction: ingress
      interfaceselector:
        interfaces:
        - eth0
      priority: 0
    type: TCX
  - tcx:
      bpffunctionname: tcx_egress_pca_parse
      direction: egress
      interfaceselector:
        interfaces:
        - eth0
      priority: 0
    type: TCX
  - fentry:
      bpffunctionname: tcp_rcv_fentry
      func_name: tcp_rcv_established
    type: Fentry
  - kprobe:
      bpffunctionname: tcp_rcv_kprobe
      func_name: tcp_rcv_established
      offset: 0
      retprobe: false
    type: Kprobe
  - kprobe:
      bpffunctionname: rh_network_events_monitoring
      func_name: rh_psample_sample_packet
      offset: 0
      retprobe: false
    type: Kprobe
  - tracepoint:
      bpffunctionname: kfree_skb
      names:
      - skb/kfree_skb
    type: Tracepoint
```

with the fix
```bash
oc get bpfprograms 
NAME                            TYPE          STATUS            AGE
netobserv-fentry-907b363d       application   bpfmanLoaded      4m59s
netobserv-kprobe-2d90081d       application   bpfmanNotLoaded   4m50s
netobserv-kprobe-edd1a87c       application   bpfmanLoaded      4m55s
netobserv-tcx-0889f5a3          application   bpfmanLoaded      5m9s
netobserv-tcx-1043c6aa          application   bpfmanLoaded      5m2s
netobserv-tcx-3c193941          application   bpfmanLoaded      5m19s
netobserv-tcx-e8296a2e          application   bpfmanLoaded      5m6s
netobserv-tracepoint-0adc0447   application   bpfmanLoaded      4m43s
```

### DIFF
--- a/controllers/bpfman-agent/application-program.go
+++ b/controllers/bpfman-agent/application-program.go
@@ -78,7 +78,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		for j, p := range a.Spec.Programs {
 			switch p.Type {
 			case bpfmaniov1alpha1.ProgTypeFentry:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fentry.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fentry.FunctionName), p.Fentry.BpfFunctionName)
 				fentryProgram := bpfmaniov1alpha1.FentryProgram{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:   buildProgramName(a, p),
@@ -100,7 +100,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				complete, res, err = r.reconcileCommon(ctx, rec, fentryObjects)
 
 			case bpfmaniov1alpha1.ProgTypeFexit:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fexit.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fexit.FunctionName), p.Fexit.BpfFunctionName)
 				fexitProgram := bpfmaniov1alpha1.FexitProgram{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:   buildProgramName(a, p),
@@ -123,7 +123,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 			case bpfmaniov1alpha1.ProgTypeKprobe,
 				bpfmaniov1alpha1.ProgTypeKretprobe:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Kprobe.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Kprobe.FunctionName), p.Kprobe.BpfFunctionName)
 				kprobeProgram := bpfmaniov1alpha1.KprobeProgram{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:   buildProgramName(a, p),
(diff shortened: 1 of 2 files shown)


anfredette (changes requested):
This looks good, but I think we have the same potential problem for all program types.  E.g., you couldn't attach two different kprobe programs to the same function.  You might not need to do this right now, but it's a possible use case, so I think we should make the same change for the appProgramId for all types.

msherif1234:
> This looks good, but I think we have the same potential problem for all program types. E.g., you couldn't attach two different kprobe programs to the same function. You might not need to do this right now, but it's a possible use case, so I think we should make the same change for the appProgramId for all types.

I don't think we can attach the same kprobe/uprobe multiple times ?

controllers/bpfman-agent/application-program.go
```diff
@@ -78,7 +78,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		for j, p := range a.Spec.Programs {
 			switch p.Type {
 			case bpfmaniov1alpha1.ProgTypeFentry:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fentry.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), p.Fentry.BpfFunctionName)
```
anfredette, line 81: I don't think we can make this same change to the types that don't support a list of attach points.  Or, if we do, we should still include the attach point itself, because if a user wants to attach the same bpf program to multiple Fentry attach points (for example), they'll need to use multiple Fentry program entries.

controllers/bpfman-agent/application-program.go
```diff
@@ -100,7 +100,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				complete, res, err = r.reconcileCommon(ctx, rec, fentryObjects)
 
 			case bpfmaniov1alpha1.ProgTypeFexit:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Fexit.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), p.Fexit.BpfFunctionName)
```
anfredette, line 103: See comment on Fentry.

controllers/bpfman-agent/application-program.go
```diff
@@ -123,7 +123,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 			case bpfmaniov1alpha1.ProgTypeKprobe,
 				bpfmaniov1alpha1.ProgTypeKretprobe:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Kprobe.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), p.Kprobe.BpfFunctionName)
```
anfredette, line 126: See comment on Fentry.

controllers/bpfman-agent/application-program.go
```diff
@@ -146,7 +146,7 @@ func (r *BpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 			case bpfmaniov1alpha1.ProgTypeUprobe,
 				bpfmaniov1alpha1.ProgTypeUretprobe:
-				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), sanitize(p.Uprobe.FunctionName))
+				appProgramId := fmt.Sprintf("%s-%s", strings.ToLower(string(p.Type)), p.Uprobe.BpfFunctionName)
```
anfredette, line 149: See comment on Fentry.

anfredette (changes requested):

controllers/bpfman-agent/application-program.go
anfredette: Thinking through this more, I think we need both the attach point and the bpf program name.

anfredette: We can get away without the attach point for the programs that support multiple attach points by mandating that people use a single entry in the bpfApplication CRD for a given program-type/bpf-program-name instead of using multiple entries of the same program type for different attach points.

anfredette (approved):
/lgtm

PR #339, opened 2024-11-05
Add sample kprobe yaml with 9 global variables

anfredette:


config/samples/bpfman.io_v1alpha1_globals_kprobeprogram.yaml
```diff
@@ -0,0 +1,39 @@
...
+    enable_dns_tracking:
+      - 0x08
+    enable_flows_filtering:
+      - 0x09
+    dns_port:
+      - 0x0A
+      - 0x0B 
+    enable_network_events_monitoring:
+      - 0x0C 
+    network_events_monitoring_groupid:
```
msherif1234, line 38: since this just sample can u use general var name instead of application specific ?

Billy99 (approved):
/lgtm

PR #342, opened 2024-11-18
API for attaching BPF programs inside network namespaces

anfredette:
This API addition allows XDP, TC, and TCX BPF programs to be attached to interfaces inside network namespaces.

The related API from the [Network Observability Operator](https://github.com/netobserv/network-observability-operator/blob/main/apis/flowcollector/v1beta2/flowcollector_types.go#L316) was adapted.

This draft API allows the interfaces and network namespaces to be wildcarded with the expectation that bfpman will automatically discover network namespaces and interfaces, but it is expected that the first implementation will require that at least the network namespace be specified.


anfredette:
I'm closing this pr and starting over with a simplified version.

PR #345, opened 2024-12-04
Add support for attaching TCX, TC and XDP programs inside containers.

anfredette:
To achieve this support, an optional container selector has been added to the TCX, TC and XDP attach info.  The agent code then finds the network namespaces of the containers and passes them to bpfman using code that already existed for attaching uprobes inside containers.

One caveat is that `GetPrimaryNodeInterface()` only finds the primary "node" interface, which may or may not be the name of the primary interface inside of a container.  As a side note, it probably works for many cases, especially when running on kind, where both the node's primary interface and containers' primary interfaces are typically named "eth0".

I propose that we leave it as-is for now, and either generalize this function to `GetPrimaryInterface()`, or add a `GetPrimaryPodInterface()` when we implement interface discovery.  We can also expand the interface selection options to include wildcards and regular expressions at that time.


Billy99 (commented):
Looks good. I'm testing now, but no real issues.

PR #347, opened 2024-12-09
bpfman-operator: Support Load/Attach Split

anfredette:
This commit introduces support for bpfman’s load-attach split from [bpfman PR
#1354](https://github.com/bpfman/bpfman/pull/1354) and includes major
refactoring and CRD changes to align with this new model.

Key Changes:

- Remove "*Program" CRDs and related code in favor of the BpfApplication CRD,
  which can contain multiple eBPF programs.
- Replace BpfProgram CRD with BpfApplicationState, which stores per-node state
  information for all eBPF programs in a BpfApplication.
- Rename CRDs for better alignment with Kubernetes conventions:
  - BpfApplication => ClusterBpfApplication (Cluster-scoped)
  - BpfNsApplication => BpfApplication (Namespace-scoped)
- The BpfApplication CRDs have been modified to include an optional list of
  attach points for each program, enabling pre-loading programs before
  attachment and supporting dynamic attachment updates.
- Redesign the bpfman-agent controller to natively support BpfApplications.

This code can be executed on KIND with the following command:
```
BPFMAN_IMG="quay.io/afredette/bpfman:latest" make run-on-kind
```

Fixes: #397


anfredette:
All of the examples will need to be updated with the new CRD format before they will pass.

anfredette:
I've cleaned up the code a bit and added support for TCX and Fentry.  See the TODO.md file for more details.

PROJECT
```diff
@@ -86,47 +23,21 @@ resources:
...
-  kind: TcNsProgram
-  path: github.com/bpfman/bpfman-operator/apis/v1alpha1
-  version: v1alpha1
-- api:
-    crdVersion: v1
-    namespaced: true
-  controller: true
-  domain: bpfman.io
-  kind: TcxNsProgram
+  kind: BpfNsApplication
```
dave-tucker, line 26: We agreed this would be `BpfApplication` and `ClusterBpfApplication`
But IIRC we wanted to get Billy's change in first before we changed the naming since this API change is the breaking one.

PROJECT
```diff
@@ -86,47 +23,21 @@ resources:
...
+  kind: BpfApplicationState
   path: github.com/bpfman/bpfman-operator/apis/v1alpha1
   version: v1alpha1
 - api:
     crdVersion: v1
-    namespaced: true
   controller: true
   domain: bpfman.io
-  kind: BpfNsApplication
+  kind: BpfNsApplicationState
```
dave-tucker, line 40: So this would then be ClusterBpfApplicationState

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
   nodeselector: {}
   bytecode:
     image:
-      url: quay.io/bpfman-bytecode/go-app-counter:latest
+      url: quay.io/bpfman-bytecode/app-test:latest
+      # imagepullpolicy: Always
   programs:
-    - type: Kprobe
+    - bpffunctionname: kprobe_test
```
dave-tucker, line 15: I really don't like `bpffunctioname`. It should either be `bpfFunctionName` or just `functionName` since we're in a `Bpf` CRD.. or hey, even just `function`

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
+      type: Kprobe
```
dave-tucker, line 16: Just wondering, is case significant here? It seems to me that `kprobe` or `KPROBE` or `Kprobe` should probably work. I'd prefer to use the lowercase version.

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
-        func_name: try_to_wake_up
-        offset: 0
-        retprobe: false
-    - type: Tracepoint
+        attach_points:
+          - func_name: try_to_wake_up
+            offset: 0
+            retprobe: false
+    - bpffunctionname: kretprobe_test
+      type: Kprobe
```
dave-tucker, line 23: Do we need to expose `type` here?
It can be inferred based on which of the optional fields are set so it seems redundant

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
+    - bpffunctionname: tc_pass_test
+      type: TC
       tc:
-        bpffunctionname: stats
-        interfaceselector:
-          primarynodeinterface: true
-        priority: 55
-        direction: ingress
-    - type: TCX
+        attach_points:
```
dave-tucker, line 37: I think this should be `links` as we're expressing desired state.
The verb might be `attach`, but we're expecting `links` in the end.

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
-    - type: Kprobe
+    - bpffunctionname: kprobe_test
+      type: Kprobe
       kprobe:
-        bpffunctionname: kprobe_counter
-        func_name: try_to_wake_up
-        offset: 0
-        retprobe: false
-    - type: Tracepoint
+        attach_points:
```
dave-tucker, line 18: 👀 inconsistent rules re: casing. k8s CRDs are supposed to be camelCase

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
+          - func_name: try_to_wake_up
```
dave-tucker, line 19: as above: `funcName`

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
+      type: TC
       tc:
-        bpffunctionname: stats
-        interfaceselector:
-          primarynodeinterface: true
-        priority: 55
-        direction: ingress
-    - type: TCX
+        attach_points:
+          - interfaceselector:
```
dave-tucker, line 38: `interfaceSelector`. I'm not going to comment on this anymore.
We seriously need to take a pass through the API to fix the inconsistencies.
We can do that in this PR, or later

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
```diff
@@ -9,50 +9,113 @@ spec:
...
+                matchLabels:
+                  name: bpfman-daemon
+              containernames:
+                - bpfman
+                - bpfman-agent
+    - bpffunctionname: fentry_test
+      type: Fentry
+      fentry:
+        function_name: do_unlinkat
+        attach: true
```
dave-tucker, line 116:
```suggestion
    - function: fentry_test
      fentry:
        links:
          - function_name: do_unlinkat
```

This shouldn't deviate from the pattern in the other program types.
The fact we need the fn_name for `load()` and we can only attach it to one place are implementation details we shouldn't be exposing in the API.

dave-tucker (changes requested):
Some feedback on the API. The naming comments we can probably handle in a follow up as this API change isn't going to be backwards compatible anyway.
We should however fix the fentry/fexit and preferably drop `type: Xdp` before merge IMHO

apis/v1alpha1/bpfApplication_types.go
```diff
@@ -42,15 +42,9 @@ const (
 	// ProgTypeKprobe refers to the Kprobe program type.
 	ProgTypeKprobe EBPFProgType = "Kprobe"
 
-	// ProgTypeKretprobe refers to the Kprobe program type.
-	ProgTypeKretprobe EBPFProgType = "Kretprobe"
-
 	// ProgTypeUprobe refers to the Uprobe program type.
```
msherif1234, line 45: I don't know why u deleted the kproberet and uproberet program types ?

apis/v1alpha1/shared_types.go
```diff
@@ -98,19 +84,58 @@ type BpfAppCommon struct {
 	// Bytecode configures where the bpf program's bytecode should be loaded
 	// from.
 	ByteCode BytecodeSelector `json:"bytecode"`
+
+	// MapOwnerSelector is used to select the loaded eBPF program this eBPF program
+	// will share a map with. The value is a label applied to the BpfProgram to select.
+	// The selector must resolve to exactly one instance of a BpfProgram on a given node
+	// or the eBPF program will not load.
```
msherif1234, line 91: we no longer have an object called `BpfProgram` so we should avoid using it everywhere in code and comments

controllers/bpfman-agent/cl-tc-program.go
```diff
@@ -0,0 +1,349 @@
...
+			attachInfo.Containers.Namespace,
+			attachInfo.Containers.Pods,
+			attachInfo.Containers.ContainerNames,
+			r.Logger,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get container pids: %v", err)
+		}
+
+		if containerInfo != nil && len(*containerInfo) != 0 {
```
msherif1234, line 298: nit no need to check the len range will handle that

msherif1234 (commented):
not sure about renaming the agent files with `cl` is that for clusterscope ?

apis/v1alpha1/bpfApplication_types.go
anfredette: They weren't being used.

anfredette:
> not sure about renaming the agent files with cl is that for clusterscope ?

Yes.  I was planning to rename things as follows:
cluster-scoped related files: cl_*
namespace-scoped related files: ns_*

PROJECT
anfredette: Done

PROJECT
anfredette: Done

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: We made it "bpffunctionname" after much discussion a while back when we started using the bpf function name instead of a name defined in the section declaration.  We also wanted to disambiguate this function name from the name of the function that some bpf programs attach to.

That said, I don't like it either :-).  I changed "bpfunctionname" to just "name", and the "func_name" and "function_name" to just "function".

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: Yes and no.  Left unchecked, I don't think Kubernetes would care, but it is case sensitive in that kubernetes passes these values through to the controller code in the case that the user enters them.  However, we also have validation logic that only allows one case, so in our case (no pun intended) it is case sensitive.  If we wanted to support case insensitivity, we'd have to handle it in our code.

I made everything lower case.  We can disuss whether to use camelCase instead.

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: This is implemented this as union, and Type is the "unionDiscriminator".  This tells us which of the "unionMembers" will be populated, and validation logic ensures that if the Type is X, the program info for X is populated.  @msherif1234 used this pattern for the original BpfApplication CRD. There may be another way to implement the API, but this works.

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: Done

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: Is the JSON representation supposed to be camelCase?

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: It's now just function

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: I've changed the pattern.

apis/v1alpha1/shared_types.go
anfredette: I got rid of a lot of it, but there's still more work to do.

controllers/bpfman-agent/cl-tc-program.go
anfredette: Done.

config/samples/bpfman.io_v1alpha1_bpfapplication.yaml
anfredette: I just went back and changed all of the json tags to camelCase

Billy99 (approved):
Work needs to be done on the integration tests. Merging with integration tests failing and work off main to get tests passing again.

PR #349, opened 2024-12-14
Add unit test for Uprobe in container

anfredette:
Add unit test for Uprobe in container

Also, move TestGetPods from uprobe-program_test.go into a separate
test_common.go file since it tests the common code used for multiple
program types.


Billy99 (approved):
/lgtm

Issue #353, opened 2024-12-23
BpfPrograms are sometimes deleted before bpfman agent can make the final status update


anfredette:
When a *Program is being deleted, the BpfPrograms are sometimes deleted before bpfman agent can make the final status update.  Everything still works, but this results in an error log message which can be alarming to users.  The solution is to do the final update before removing the finalizer.

PR #354, opened 2024-12-23
Fix intermittent error when deleting programs

anfredette:
Change the order of operations in `handleProgDelete()` to update the status before removing the finalizer when a *Program is being deleted.

Fixes: #353


Billy99 (approved):
/lgtm

PR #356, opened 2025-01-02
Update go version to 1.23

anfredette:


anfredette:
We're not going to make this change just yet.

PR #362, opened 2025-01-17
update golang to 1.23 and build crictl from source

msherif1234:
bump go to 1.23 to reenable dependabot latest updates and cves fixes
and build crictl from source


anfredette (approved):
/LGTM

PR #371, opened 2025-02-11
Fix random failures in unit tests

anfredette:
Failures were caused when the same randomly generated id was used for more than one program
in the fake bpfman client. Changed the code to use a deterministic incrementing ID.

### DIFF
--- a/controllers/bpfman-agent/internal/test-utils/fake_bpfman_client.go
+++ b/controllers/bpfman-agent/internal/test-utils/fake_bpfman_client.go
@@ -19,7 +19,6 @@ package testutils
 import (
 	"context"
 	"fmt"
-	"math/rand"
 
 	gobpfman "github.com/bpfman/bpfman/clients/gobpfman/v1"
 	grpc "google.golang.org/grpc"
@@ -55,8 +54,12 @@ func NewBpfmanClientFakeWithPrograms(programs map[int]*gobpfman.ListResponse_Lis
 	}
 }
 
+var currentID = 1000
+
 func (b *BpfmanClientFake) Load(ctx context.Context, in *gobpfman.LoadRequest, opts ...grpc.CallOption) (*gobpfman.LoadResponse, error) {
-	id := rand.Intn(100)
+	currentID++
+	id := currentID
+
 	b.LoadRequests[id] = in
 
 	b.Programs[id] = loadRequestToListResult(in, uint32(id))


controllers/bpfman-agent/internal/test-utils/fake_bpfman_client.go
```diff
@@ -56,7 +56,14 @@ func NewBpfmanClientFakeWithPrograms(programs map[int]*gobpfman.ListResponse_Lis
...
 
 func (b *BpfmanClientFake) Load(ctx context.Context, in *gobpfman.LoadRequest, opts ...grpc.CallOption) (*gobpfman.LoadResponse, error) {
-	id := rand.Intn(100)
+	var id int
+	for {
+		id = rand.Intn(30000)
+		if _, ok := b.LoadRequests[id]; !ok {
+			break
+		}
+	}
```
msherif1234, line 65: can u avoid infinite loop and just have maxRetries and fail if its reached ?

anfredette: I figured someone wouldn't like the infinite loop :-).  How about this new solution that doesn't use random numbers?

controllers/bpfman-agent/internal/test-utils/fake_bpfman_client.go
```diff
@@ -55,8 +54,12 @@ func NewBpfmanClientFakeWithPrograms(programs map[int]*gobpfman.ListResponse_Lis
 	}
 }
 
+var currentID = 1000
+
 func (b *BpfmanClientFake) Load(ctx context.Context, in *gobpfman.LoadRequest, opts ...grpc.CallOption) (*gobpfman.LoadResponse, error) {
-	id := rand.Intn(100)
+	currentID++
+	id := currentID
```
msherif1234, line 61: do u still need currentID ?

anfredette: BTW, I've seen this issue for a long time.  I.e., unit tests would fail sometimes, but when you run them again they would pass.  It just started happening often enough when I started using more programs that it bothered me enough to figure out why.

msherif1234: i was just thinking of adding a cntr of 100 or 100 retries to get unused id from random instead of infinite loop

anfredette: I understood what you were suggesting, and I think it would work, but I think this is better -- it's simpler and always unique because you remember the last ID and increment it for the next one.

I thought about doing it this way after I made the first change, but figured the previous way was good enough and didn't bother going back to change it.

msherif1234: sure I am fine either way was just answering ur question

msherif1234 (approved):

PR #372, opened 2025-02-11
Add "patch" to verbs for namespace-scoped CRDs in namespace_scoped.yaml

anfredette:
This allows the user to update the CRDs.

### DIFF
--- a/hack/namespace_scoped.yaml
+++ b/hack/namespace_scoped.yaml
@@ -42,7 +42,6 @@ rules:
     - watch
 - apiGroups:
     - bpfman.io
-  # resources: ['xdpnsprograms']
   resources: ['bpfnsapplications', 'tcnsprograms', 'tcxnsprograms', 'uprobensprograms', 'xdpnsprograms']
   verbs:
     - create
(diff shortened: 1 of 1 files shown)


msherif1234:
don't u need to add it
https://github.com/bpfman/bpfman-operator/blob/main/config/rbac/bpfman-agent/role.yaml
and
https://github.com/bpfman/bpfman-operator/blob/main/config/rbac/bpfman-operator/role.yaml

then run 
`make manifests bundle`

anfredette:
> don't u need to add it https://github.com/bpfman/bpfman-operator/blob/main/config/rbac/bpfman-agent/role.yaml and https://github.com/bpfman/bpfman-operator/blob/main/config/rbac/bpfman-operator/role.yaml
> 
> then run `make manifests bundle`

I could be wrong, but I didn't think so.  This file is used for testing the namespace-scoped CRDs.  It's used to create a ServiceAccount that can apply the namespace-scoped CRDs in a particular namespace.  Billy has some instructions for using it, but I don't know if they've been pushed or merged yet.  

@Billy99 what do you think?

Billy99:
Here are the docs for it:
https://bpfman.io/main/getting-started/operator-quick-start/#deploy-namespace-scoped-sample

Billy99 (approved):
/lgtm

PR #378, opened 2025-02-18
rename bundle containerfile to Containerfile.bundle to be consistent

msherif1234:
rename bundle container file to be consistent with other container files in the repo


Billy99:
Do you need to update Container.bundle.openshift with sdk 1.27.0?

msherif1234:
> Do you need to update Container.bundle.openshift with sdk 1.27.0?

probably at some point but not urgent

Billy99 (approved):
/lgtm

PR #381, opened 2025-02-21
feat: add slack notification on failed integration tests

### LINKED ISSUE #332: Notify Kubernetes #bpfman channel if nightly bpfman-operator integration test fails

hanshal101:
Fixes #332 
Closes #332 

### Description
This PR adds the feature to give alerts on slack when the integration tests fails.

### TODO
@anfredette , we need to add a Github secret named `SLACK_WEBHOOK_URL` to enable the alerts.

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -80,3 +80,10 @@ jobs:
           name: kubernetes-integration-test-diag
           path: /tmp/ktf-diag*
           if-no-files-found: ignore
+
+      - name: Notify Slack on Test Failure
+        if: failure()
+        uses: rtCamp/action-slack-notify@v2
+        with:
+          webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          message: "Nightly Kubernetes BPFMan Operator integration tests failed! Go version: ${{ matrix.go }} and OCI runtime: ${{ matrix.oci_bin }}"
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Check disk space
         run: df -h
 
-      ## Upload diagnostics if integration test step failed.
+      # Upload diagnostics if integration test step failed.
       - name: Upload diagnostics
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yml
```diff
@@ -80,3 +80,10 @@ jobs:
           name: kubernetes-integration-test-diag
           path: /tmp/ktf-diag*
           if-no-files-found: ignore
+
+      - name: Notify Slack on Test Failure
+        if: failure()
+        uses: rtCamp/action-slack-notify@v2
```
dave-tucker, line 86: It might be better to use the action supported by Slack themselves.
There's a fancy example here: https://github.com/slackapi/slack-github-action/blob/main/example-workflows/Technique_3_Slack_Incoming_Webhook/blocks.yml
And a much simpler one here: https://github.com/slackapi/slack-github-action/blob/main/example-workflows/Technique_3_Slack_Incoming_Webhook/text.yml

.github/workflows/integration_test.yml
```diff
@@ -80,3 +80,10 @@ jobs:
...
+        with:
+          webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          message: "Nightly Kubernetes BPFMan Operator integration tests failed! Go version: ${{ matrix.go }} and OCI runtime: ${{ matrix.oci_bin }}"
```
dave-tucker, line 89:
```suggestion
          message: "Nightly Kubernetes bpfman Operator integration tests failed! Go version: ${{ matrix.go }} and OCI runtime: ${{ matrix.oci_bin }}"
```

dave-tucker (changes requested):
Thanks for the contribution. I left some comments, but reading up on this some more it seems like webhooks are ~being deprecated~ no longer recommended for this usecase + we don't have the ability to add them easily on the K8s slack instance. 

We can certainly request a webhook from the admins, but it might make more sense to do use the Github integration like this: https://github.com/integrations/slack?tab=readme-ov-file#workflow-notification-filters

That's more flexible since we can subscribe/unsubscribe to notifications in the channel whenever we like.

hanshal101:
Thanks @dave-tucker for the review, I have updated the CI to the official `slackapi`, I have added a simple string for the message, let me know if we need more detailed one or not.

Issue #399, opened 2025-03-03
Make the bpfman-operator report status of "Pending" instead of "Error" when appropriate


anfredette:
If one or more BpfApplicationState objects are in the "Pending" state, but none are in "Error" state, we should report the BpfApplication state as "Pending".

PR #409, opened 2025-03-06
Make the bpfman-operator report status of "Pending" instead of "Error" when appropriate

anfredette:
If any nodes report an error, the operator will also report an error. However, if there are no errors but some nodes are in a pending state, the operator will report as pending.

Fixes: #399

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -154,28 +154,6 @@ func IsBpfmanDeployed() bool {
 	return false
 }
 
-func IsBpfApplicationConditionFailure(conditions *[]metav1.Condition) bool {
-	if conditions == nil || *conditions == nil || len(*conditions) == 0 {
-		return true
-	}
-
-	numConditions := len(*conditions)
-
-	if numConditions > 1 {
-		// We should only ever have one condition so log a message, but
-		// still look at (*conditions)[0].
-		log.Info("more than one condition found", "numConditions", numConditions)
-	}
-
-	if (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondPending) ||
-		(*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondError) ||
-		(*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondProgramListChangedError) {
-		return true
-	}
-
-	return false
-}
-
 func IsBpfAppStateConditionFailure(conditions *[]metav1.Condition) bool {
 	if conditions == nil || *conditions == nil || len(*conditions) == 0 {
 		return true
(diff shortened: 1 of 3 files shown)


pkg/helpers/helpers.go
```diff
@@ -189,9 +189,9 @@ func IsBpfAppStateConditionFailure(conditions *[]metav1.Condition) bool {
 		log.Info("more than one condition found", "numConditions", numConditions)
 	}
 
-	if (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppCondSuccess) {
-		return false
-	} else {
+	if (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppCondPending) {
 		return true
+	} else {
```
msherif1234, line 194: no need for else here and the same for failed

anfredette: Done

pkg/helpers/helpers.go
```diff
@@ -167,16 +167,12 @@ func IsBpfApplicationConditionFailure(conditions *[]metav1.Condition) bool {
...
 	}
 
-	if (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondPending) ||
-		(*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondError) ||
-		(*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondProgramListChangedError) {
-		return true
-	}
-
-	return false
+	return (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppStateCondError) ||
```
msherif1234, line 170: r we sure condition[0] will always have error in this case can condition[0] be success but [1] is the failing one ? maybe we can just loop and if any of the condition match what u have we return it ?

pkg/helpers/helpers.go
```diff
@@ -189,9 +185,5 @@ func IsBpfAppStateConditionFailure(conditions *[]metav1.Condition) bool {
 		log.Info("more than one condition found", "numConditions", numConditions)
 	}
 
-	if (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppCondSuccess) {
-		return false
-	} else {
-		return true
-	}
+	return (*conditions)[0].Type == string(bpfmaniov1alpha1.BpfAppCondPending)
```
msherif1234, line 188: same here does [0] gruanteed to be in pending state or we better off loop through all and return the 1st one in pending state ?

pkg/helpers/helpers.go
anfredette: The way we use conditions is wonky and we should fix it, but as it is, there should only ever be 1 condition and the actual condition should always be at index 0.  See https://github.com/bpfman/bpfman-operator/blob/main/controllers/bpfman-agent/common.go#L160 for the details.  It's an internal bug if `numConditions > 1` here, so we log it, but it shouldn't ever happen with the current code.

msherif1234 (approved):

PR #416, opened 2025-03-18
revert back to go1.23 and pin cri-tool to v1.32.0

msherif1234:
this PR we have to pin cri-tool to version `v1.32.0` to avoid bump to go1.24


Billy99 (commented):
/lgtm
@anfredette I think you added cri-tools dependency, assume you are good with pining it?

PR #421, opened 2025-03-26
Add intefaces match and regex filter to autodiscovery configs

msherif1234:
with modified intf discovery yaml we see TC only attached to `veth` interfaces and not to `eth0`

```yaml
 programs:
    - name: tc_pass_test
      programId: 5443
      programLinkStatus: Success
      tc:
        links:
        - direction: Ingress
          interfaceName: veth4fbd563a
          linkId: 2420163623
          linkStatus: Attached
          priority: 0
          proceedOn:
          - Pipe
          - DispatcherReturn
          shouldAttach: true
          uuid: d106177e-a1e5-46e4-aae2-3b578567a77e
        - direction: Ingress
          interfaceName: vethf1705d1a
          linkId: 282060526
          linkStatus: Attached
          priority: 0
          proceedOn:
          - Pipe
          - DispatcherReturn
          shouldAttach: true
          uuid: 105a8a63-ff61-443f-929d-0af77adaa09b
        - direction: Ingress
          interfaceName: veth7bb5fd13
          linkId: 808594878
          linkStatus: Attached
          priority: 0
          proceedOn:
          - Pipe
          - DispatcherReturn
          shouldAttach: true
          uuid: c5c93069-0585-4592-8c44-6deb452e5005
        - direction: Ingress
          interfaceName: vethf38ce0a8
          linkId: 4238487788
          linkStatus: Attached
          priority: 0
          proceedOn:
          - Pipe
          - DispatcherReturn
          shouldAttach: true
          uuid: 070c59a8-7288-43c4-b747-eba8703c7d73
      type: TC
```


controllers/bpfman-agent/common.go
```diff
@@ -240,15 +242,57 @@ func interfaceInExcludeList(interfaceSelector *bpfmaniov1alpha1.InterfaceSelecto
...
+		}
+	}
+	return false
+}
+
 func getInterfaces(interfaceSelector *bpfmaniov1alpha1.InterfaceSelector, ourNode *v1.Node, discoveredInterfaces *sync.Map) ([]string, error) {
 	var interfaces []string
 
 	if interfaceSelector.InterfacesDiscoveryConfig != nil && interfaceSelector.InterfacesDiscoveryConfig.InterfaceAutoDiscovery != nil &&
 		*interfaceSelector.InterfacesDiscoveryConfig.InterfaceAutoDiscovery {
```
anfredette, line 290: Your changes in this PR lgtm.  However, looking at these lines makes me wonder whether we need `InterfaceAutoDiscovery *bool`.  It seems like we can do without it and the fact that `interfaceSelector.InterfacesDiscoveryConfig != nil` would tell us that it's enabled.  What do you think?

anfredette (commented):
These changes lgtm, but please see my comment about the existing code.  Thanks, Andre

controllers/bpfman-agent/common.go
msherif1234: what if both lists are nil like excludeInterfaces=nil and AllowedInterfaces=nil and we need to still do autodiscovery won't the bool here will be handy ?

msherif1234: if we don't provide anything i.e. have InterfaceDiscoveryConfig we will get validation error 
`* spec.programs[1].tcx.links[0].interfaceSelector: Invalid value: 0: spec.programs[1].tcx.links[0].interfaceSelector in body should have at least 1 properties`
so we need to have one field populated

controllers/bpfman-agent/common.go
```diff
@@ -240,15 +243,56 @@ func interfaceInExcludeList(interfaceSelector *bpfmaniov1alpha1.InterfaceSelecto
...
+
+	for _, n := range allowedMatches {
+		if n == intf {
+			return true
+		}
+	}
+	return false
+}
+
 func getInterfaces(interfaceSelector *bpfmaniov1alpha1.InterfaceSelector, ourNode *v1.Node, discoveredInterfaces *sync.Map) ([]string, error) {
```
anfredette, line 287: I was thinking we could define something like:
```
type netnsInterface struct {
	interfaceName string
	netnsPath     string
}
```
And then have `getInterfaces()` return `[]netnsInterface`.  The code that handles interface discovery could populate netnsPath as appropriate as it goes through the loop.  If an interface isn't in a network namespace, then we could set `netnsPath` to `""`.  This would be the case for all of the interfaces returned if interface discovery isn't enabled.  What do you think?

controllers/bpfman-agent/cl_tc_program.go
```diff
@@ -322,19 +322,39 @@ func (r *ClTcProgramReconciler) getExpectedLinks(ctx context.Context, attachInfo
...
-					ShouldAttach: true,
-					UUID:         uuid.New().String(),
-					LinkId:       nil,
-					LinkStatus:   bpfmaniov1alpha1.ApAttachNotAttached,
-				},
-				InterfaceName: iface,
-				Priority:      attachInfo.Priority,
-				Direction:     attachInfo.Direction,
-				ProceedOn:     attachInfo.ProceedOn,
+			netnsList := getInterfaceNetNsList(&attachInfo.InterfaceSelector, iface, r.Interfaces)
```
anfredette, line 325: I could be misunderstanding something, but it doesn't look like this code handles the case where an interface (e.g., eth0) is discovered both in a network namespace and in the root network namespace (unless you're including the netnsPath for the root namespace too).

controllers/bpfman-agent/common.go
anfredette: Just thinking some more, we might want to have a different function for this calle `getDiscoveredInterfaces()` for this, and have `getInterfaces()` still return `[]string`.

anfredette (approved):
I have a question about whether this works when we have interface discovery supported inside network namespaces, and I think it can be made more efficient as described in comments.  However, it works, so I'm going to approve it as-is and we can handle these concerns as in a follow-up pr.

PR #423, opened 2025-03-31
Move BpfApplicationState Spec to Status

anfredette:
To do this, I had to rework the reconcile functions a bit.

Also:
- Replaced the hand-coded wait loops with wait.PollUntilContextTimeout() from k8s.io/apimachinery.
- Added a bpfapplication sample that attaches to the test-target deployment because I almost always get ErrImagePull when trying to deploy nginx on my lab machine.


controllers/bpfman-agent/cl_application_program.go
```diff
@@ -156,56 +150,32 @@ func (r *ClBpfApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+				// If the BpfApplicationState doesn't exist and the BpfApplication
+				// is being deleted, we don't need to do anything.  Just continue
+				// with the next BpfApplication.
+				r.Logger.Info("BpfApplicationState doesn't exist and BpfApplication is being deleted",
+					"Name", r.currentApp.Name)
+				continue
+			}
+			// Create a new BpfApplicationState object first, once it's created,
+			// initialize the Status subresource and then update the status.
+			return r.createBpfAppState(ctx)
```
msherif1234, line 171: nit: comments should say ClBpfApplicationState

controllers/bpfman-agent/cl_application_program.go
```diff
@@ -441,115 +427,99 @@ func (r *ClBpfApplicationReconciler) getProgState(prog *bpfmaniov1alpha1.ClBpfAp
...
-				"currentGeneration", bpfAppState.GetGeneration())
-			r.currentAppState = bpfAppState
-			return nil
-		}
-		time.Sleep(retryInterval)
+	r.Logger.V(1).Info("Start waitForBpfAppStateStatusUpdate()")
+	i := 0
+	err := wait.PollUntilContextTimeout(ctx, updateRetryInterval, updateTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			i++
```
msherif1234, line 499: I don't think we need to use `i` anymore in here

anfredette: You are correct that we don't need it from a functional perspective, however, I'm currently logging a message that says how many iterations it took to get the updated CRD so we can track how it works on different/larger systems.  And, I'm not aware of another way to know how many polls were required.

msherif1234: normally we don't track iterations just log when timeout condition happens tracking iterations might jist generate noise in the logs

anfredette: I think we can eventually get rid of this, but we weren't sure what was the right amount of time to wait.  So, I thought it would be good to do some more testing on larger deployments, and see exactly  how long it's taking for the updates to become available.

msherif1234 (approved):

controllers/bpfman-agent/cl_application_program.go
anfredette: Done

controllers/bpfman-operator/common_cluster.go
```diff
@@ -74,7 +74,7 @@ func statusChangedPredicateCluster() predicate.Funcs {
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			oldObject := e.ObjectOld.(*bpfmaniov1alpha1.ClusterBpfApplicationState)
 			newObject := e.ObjectNew.(*bpfmaniov1alpha1.ClusterBpfApplicationState)
-			return !reflect.DeepEqual(oldObject.GetStatus(), newObject.Status)
+			return !reflect.DeepEqual(oldObject.GetStatus(), newObject.Status.Conditions)
```
Billy99, line 77: Need a similar change in controllers/bpfman-operator/common_namespace.go.

anfredette: That may have been why the integration test cleanup was failing.

apis/v1alpha1/bpf_application_state_types.go
```diff
@@ -123,8 +131,8 @@ func (an BpfApplicationState) GetLabels() map[string]string {
 	return an.Labels
 }
 
-func (an BpfApplicationState) GetStatus() *BpfAppStatus {
-	return &an.Status
+func (an BpfApplicationState) GetStatus() []metav1.Condition {
```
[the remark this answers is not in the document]
anfredette: I made similar changes in the agent controller code -- changing "status" to "condition" where appropriate.

msherif1234 (approved):
/LGTM

PR #424, opened 2025-03-31
split out interface discovery processing

msherif1234:
Address review comment https://github.com/bpfman/bpfman-operator/pull/421#discussion_r2017527615


controllers/bpfman-agent/cl_tc_program.go
```diff
@@ -290,72 +317,26 @@ func (r *ClTcProgramReconciler) getExpectedLinks(ctx context.Context, attachInfo
...
-						ProceedOn:     attachInfo.ProceedOn,
-					}
-					nodeLinks = append(nodeLinks, link)
-				}
-			}
-		} else {
-			// This is an error -- either namespacePath or pods must be set.
-			r.Logger.Error(fmt.Errorf("neither namespacePath nor pods is set"), "internal error")
+		if containerInfo == nil {
+			return nil, fmt.Errorf("neither namespacePath nor pods is set in NetworkNamespaces")
```
anfredette, line 324: I was previously logging an error, but this isn't really an error.  It's possible for the user to include a namespace selector, but none are found on the given node.  For example, maybe the pod they want to attach in didn't get scheduled on the node, or maybe it hasn't been deployed yet.

anfredette (changes requested):
I just have the one comment that applies to all the cl program types.  Otherwise, this looks good to me!  Thanks, Andre

controllers/bpfman-agent/cl_tc_program.go
msherif1234: I put the error log again but I don't think its a good idea to crash if containerinfo is NULL we probably should ensure that in the API so that will never happen

anfredette: Sorry.  I wasn't clear on that log.  That condition isn't an error, and I shouldn't have had an error log.  I think that log was originally in a different place and happened if here wasn't a container selector and there weren't any interfaces either.  The API validation should have prevented it, so it shouldn't have ever happened, but I added the log anyway.  However, I think that as the code was refactored, the error log got moved into a place where it shouldn't have been.  

That said, before your current change, if it happened that there weren't any containers found, it would have just dropped through and returned an empty `nodeLinks` slice which is the intended behavior.  However, since you moved the check up, and didn't do a return there, you are correct that the code continues and would crash if `containerInfo != nil`.  I actually had the checks in the original order (i.e., `containerInfo != nil`) so it would fall through and return the empty  `nodeLinks` slice as intended.  So, I think you can either keep it the way it is and do a `return nodeLinks, nil` after the `if containerInfo == nil` check, or swap the order back.

controllers/bpfman-agent/cl_tc_program.go
```diff
@@ -290,72 +317,28 @@ func (r *ClTcProgramReconciler) getExpectedLinks(ctx context.Context, attachInfo
...
-						ProceedOn:     attachInfo.ProceedOn,
-					}
-					nodeLinks = append(nodeLinks, link)
-				}
-			}
-		} else {
+		if containerInfo == nil {
 			// This is an error -- either namespacePath or pods must be set.
 			r.Logger.Error(fmt.Errorf("neither namespacePath nor pods is set"), "internal error")
+			return nodeLinks, nil
```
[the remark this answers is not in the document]
anfredette: And, delete the comment because it's incorrect.

anfredette (approved):
/LGTM

PR #426, opened 2025-04-05
Check for Duplicate Network Namespace Inodes

anfredette:
It’s possible for multiple pods to share the same network namespace, even though unique handles for that namespace are created in /var/run/netns. This most commonly occurs when pods are configured with "hostNetwork": true.

The interface discovery process watches /var/run/netns, and when it detects a new entry, it begins monitoring the interfaces in that namespace and informs bpfman. However, if bpfman attempts to attach multiple programs — especially TCX programs — to the same interface in the same netns, it will return an error. It's also possible to run into this situation using the network namespace selector.

For TC and XDP programs, the attach might not fail immediately, but bpfman limits the number of programs per attach point (currently to 10). If more than 10 handles reference the same network namespace — which is common on OpenShift clusters — this limit is quickly reached. That can lead to subtle and unexpected failures, even though the individual netns paths appear different.

The solution is to compare inodes rather than file paths when identifying network namespaces. This allows the system to correctly recognize when multiple entries actually refer to the same underlying namespace.

Additional changes:
- bpfman needs to mount /proc to get the inodes for interfaces inside
  pods.
- getDiscoveredInterfaces() shouldn't return an error if no interfaces
  are found.  This could be valid depending on the configuration of
  allowed and excluded interfaces.
- Added a NetnsCache to remember mapping the results of inode lookups.
  This avoids multiple syscalls for the same network namespace path.
  This cache is initialized at the beginning of each Reconcile loop
  so that changes can be detected.

Fixes: #433

### DIFF
--- a/controllers/bpfman-agent/cl_xdp_program.go
+++ b/controllers/bpfman-agent/cl_xdp_program.go
@@ -187,8 +187,8 @@ func (r *ClXdpProgramReconciler) findLink(attachInfoState bpfmaniov1alpha1.ClXdp
 		// attachInfoState is the same as a if the the following fields are the
 		// same: IfName, NetnsPath, Priority, and ProceedOn.
 		if a.InterfaceName == attachInfoState.InterfaceName && a.Priority == attachInfoState.Priority &&
-			reflect.DeepEqual(a.NetnsPath, attachInfoState.NetnsPath) &&
-			reflect.DeepEqual(a.ProceedOn, attachInfoState.ProceedOn) {
+			reflect.DeepEqual(a.ProceedOn, attachInfoState.ProceedOn) &&
+			reflect.DeepEqual(getInode(r.Logger, a.NetnsPath), getInode(r.Logger, attachInfoState.NetnsPath)) {
 			return &i
 		}
 	}
--- a/controllers/bpfman-agent/cl_xdp_program.go
+++ b/controllers/bpfman-agent/cl_xdp_program.go
@@ -162,7 +162,11 @@ func (r *ClXdpProgramReconciler) updateLinks(ctx context.Context, isBeingDeleted
 				return fmt.Errorf("failed to get node links: %v", error)
 			}
 			for _, link := range expectedLinks {
-				index := r.findLink(link)
+				index, err := r.findLink(link)
+				if err != nil {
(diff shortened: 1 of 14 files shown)


controllers/bpfman-agent/cl_xdp_program.go
```diff
@@ -187,8 +187,8 @@ func (r *ClXdpProgramReconciler) findLink(attachInfoState bpfmaniov1alpha1.ClXdp
 		// attachInfoState is the same as a if the the following fields are the
 		// same: IfName, NetnsPath, Priority, and ProceedOn.
 		if a.InterfaceName == attachInfoState.InterfaceName && a.Priority == attachInfoState.Priority &&
-			reflect.DeepEqual(a.NetnsPath, attachInfoState.NetnsPath) &&
-			reflect.DeepEqual(a.ProceedOn, attachInfoState.ProceedOn) {
+			reflect.DeepEqual(a.ProceedOn, attachInfoState.ProceedOn) &&
+			reflect.DeepEqual(getInode(r.Logger, a.NetnsPath), getInode(r.Logger, attachInfoState.NetnsPath)) {
```
msherif1234, line 191: can't we use nsHandle here as it contains Inode info already
https://github.com/vishvananda/netns/blob/v0.0.5/nshandle_linux.go#L39 
so we can avoid doing another sysCall ?

anfredette: I was thinking about that.  I think we'd end up needing to add the inode to the attach info in the crd for comparisons with other interfaces.  That would be more efficient, but take up more space in the crd.  We also still need to do this check for namespaces that didn't come from interface discovery.  I guess we could have code that only does the lookup when we didn't already get it from the nsHandle.  Another issue I've run into is that I've seen new interfaces in the discovered interface list, with network namespaces that don't exist.  It's possible that they had been created but then deleted.  They should get removed from the discovered list, but they aren't.  In this case, my new code (that I haven't pushed yet) logs an error, and doesn't try to do the attach (which would fail).  We should try to figure out why this is happening, but for now at least bpfman works.  So, my thinking is that we should keep this option on the back burner and treat it as a possible future performance optimization at the cost of extra info saved.

anfredette: Added a NetnsCache (described above) to make this more efficient.

dave-tucker (approved):

PR #449, opened 2025-07-01
Samples: Fix XDP program URL

andreaskaris:
Samples: Fix XDP program URL


dave-tucker:
What issue is this fixing? The image in that sample is valid, but you can only see it when you are logged into quay.io.
ISTR it was previously used for testing, pulling when using private images.

Issue #468, opened 2025-08-22
bpfman-operator has badly configured bidirectional mounts for the csi driver that can make system unresponsive


andreaskaris:
The bpfman-operator has badly configured bidirectional mounts for the csi driver.
While testing, I consistently managed to make my node unresponsive due to this

```
volumeMounts:
    - mountPath: /run/bpfman-sock
      name: bpfman-sock
    - mountPath: /run/bpfman
      mountPropagation: Bidirectional
      name: runtime
    - mountPath: /sys/kernel/debug
      name: host-debug
    - mountPath: /var/lib/bpfman
      mountPropagation: Bidirectional
      name: bpfman-content-store
    - mountPath: /sys/fs/bpf
      mountPropagation: Bidirectional
      name: default-bpf-fs
    - mountPath: /run/bpfman/csi
      name: socket-dir
```

This leads to tons of mounts being created and propagated to the host system when restarting the container:

```
[root@sno10-1 ~]# mount | grep bpf
bpf on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
bpffs on /sys/fs/bpf type bpf (rw,relatime)
/dev/sdb4 on /run/bpfman/csi type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /run/bpfman/csi type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /var/lib/kubelet/plugins/csi-bpfman type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /run/bpfman/csi type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /var/lib/kubelet/plugins/csi-bpfman type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /var/lib/kubelet/plugins/csi-bpfman type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdb4 on /run/bpfman/csi type xfs (rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
[...]

PR #469, opened 2025-08-22
Daemonset: Move bidirectional mount from /run/bpfman to /run/bpfman/csi

andreaskaris:
The bidirectional mount in /run/bpfman caused issues because /run/bpfman/csi was mounted inside there, resulting in an *exponentially* increasing number of mounts on the host system (double the number of mounts for every restart of the container). Moving the bidirectional statement from /run/bpfman to /run/bpfman/csi should fix this while preserving the requirement for bidirectional mounts for CSI.

Reported-at: https://github.com/bpfman/bpfman-operator/issues/468

### DIFF
--- a/config/bpfman-deployment/daemonset.yaml
+++ b/config/bpfman-deployment/daemonset.yaml
@@ -78,7 +78,6 @@ spec:
               mountPath: /run/bpfman-sock
             - name: runtime
               mountPath: /run/bpfman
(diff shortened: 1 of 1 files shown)


frobware:
I've been running deploy/undeploy for close to an hour - KIND/OpenShift clusters. This seems to fix the issue nicely. Thanks!

LGTM

PR #461, opened 2025-08-11
bpfman-operator: Add Config resource and update reconciler and tests 

andreaskaris:
## Summary

This PR introduces a new Config CRD to replace ConfigMap usage for managing cluster-scoped resources, addressing ownership reference limitations where namespace-scoped ConfigMaps cannot own cluster-scoped resources like CSIDriver. The change includes lifecycle testing and reconciliation improvements.

## Changes

- **Add Config CRD**: New custom resource definition with proper API types and validation
- **Replace ConfigMap controller**: Migrate from configmap.go to config.go with enhanced reconciliation logic
- **Fix ownership references**: Enable proper cascading deletion for cluster-scoped resources
- **Add finalizer protection**: Prevent race conditions during Config deletion with proper finalizer management
- **Improve reconciliation**: Use semantic comparison to prevent unnecessary resource updates and reconciliation loops
- **Enhanced error handling**: Add structured logging and better error reporting in reconciler
- **Update deployment config**: Modify bpfman-deployment configuration to use new Config resource
- **Add lifecycle testing**: Comprehensive integration test suite covering resource creation, updates, and restoration scenarios
- **Refactor OpenShift detection**: Extract OpenShift detection logic to internal package for better testability
- **Generate client code**: Auto-generated deepcopy, client, and informer code for Config resource
- **Sync deployment manifests**: Update DaemonSet and metrics-proxy YAML with missing Kubernetes default fields

## Technical Details

The transition from ConfigMap to Config CRD resolves the fundamental issue where namespace-scoped resources cannot establish  owner references with cluster-scoped resources. The new Config controller provides the same functionality with proper resource  lifecycle management through Kubernetes garbage collection.

Key improvements include:
- **Finalizer-based deletion**: Ensures coordinated cleanup preventing race conditions between operator and garbage collection
- **Semantic comparison**: Prevents reconciliation loops by only updating resources when meaningful changes occur
- **Field validation**: API server validation for configuration values through CRD schema
- **Lifecycle testing**: End-to-end test coverage for Config resource operations

The finalizer ensures a safe deletion sequence: user deletes Config → deletion timestamp set → operator stops creating/updating resources → finalizer removed → garbage collection proceeds via owner references.

## Testing

All existing functionality is preserved with enhanced test coverage for edge cases and error scenarios.

Tested with: (/)

* ci integration test:
```
[...]

PR #475, opened 2025-09-10
Config follow-up: Implement component override functionality for Config resource

andreaskaris:
This commit adds the ability to mark components as unmanaged in the
bpfman-operator, preventing the operator from creating or updating
specific objects. The implementation includes:

- Added ComponentOverride struct to Config API with fields for
kind, group, namespace, name, and unmanaged flag
- Modified assureResource function to check for overrides and skip
management of unmanaged components
- Implemented isOverridden helper function to match objects against
override specifications
- Added tests to verify override functionality works correctly
across all component types

### Component Overrides

The `overrides` field allows you to mark specific components as unmanaged, preventing the operator from creating or updating them. This is intended as a debug/escape hatch for advanced users who need to customize specific resources without operator interference.

#### YAML Configuration

Add overrides directly to the Config resource:

```yaml
kind: Config
metadata:
  name: bpfman-config
spec:
  # Existing configuration...
  namespace: bpfman
  daemon:
    image: quay.io/bpfman/bpfman:latest
    logLevel: info
  agent:
    image: quay.io/bpfman/bpfman-agent:latest
    logLevel: info
  configuration: |
    # bpfman.toml content here

  # Component override for the bpfman daemon DaemonSet
  overrides:
  - apiVersion: apps/v1
    kind: DaemonSet
    namespace: bpfman
    name: bpfman-daemon
    unmanaged: true
```

#### JSON Patch for Precise Control

Using `kubectl patch` to add overrides:

```bash
kubectl patch config bpfman-config --type='json' -p='[
  {
    "op": "add",
    "path": "/spec/overrides",
    "value": [
      {
        "apiVersion": "apps/v1",
        "kind": "DaemonSet",
        "namespace": "bpfman",
        "name": "bpfman-daemon",
        "unmanaged": true
      }
    ]
  }
]'
```

#### Adding Multiple Overrides

To append to existing overrides without replacing them:

```bash
kubectl patch config bpfman-config --type='json' -p='[
  {
    "op": "add",
    "path": "/spec/overrides/-",
    "value": {
      "apiVersion": "apps/v1",
      "kind": "DaemonSet",
      "namespace": "bpfman",
      "name": "bpfman-daemon",
      "unmanaged": true
    }
  }
]'
```

The `/-` notation appends to the existing overrides array rather than replacing it.

#### Examples

**Override the metrics proxy DaemonSet:**
```yaml
overrides:
- apiVersion: apps/v1
  kind: DaemonSet
  namespace: bpfman
  name: bpfman-metrics-proxy
  unmanaged: true
```

**Override a ConfigMap:**
```yaml
overrides:
- apiVersion: v1
  kind: ConfigMap
  namespace: bpfman
  name: bpfman-config
  unmanaged: true
```

**Override multiple resources:**
```yaml
overrides:
- apiVersion: apps/v1
  kind: DaemonSet
  namespace: bpfman
  name: bpfman-daemon
  unmanaged: true
- apiVersion: apps/v1
  kind: DaemonSet
  namespace: bpfman
  name: bpfman-metrics-proxy
  unmanaged: true
- apiVersion: storage.k8s.io/v1
  kind: CSIDriver
  namespace: ""  # cluster-scoped resource
  name: csi.bpfman.io
  unmanaged: true
```

**Note:** The `apiVersion` field matches what you see in kubectl output (e.g., `apps/v1` for DaemonSets, `v1` for core resources like ConfigMaps), making it easy to copy/paste from existing manifests.

### DIFF
--- a/controllers/bpfman-operator/config.go
+++ b/controllers/bpfman-operator/config.go
@@ -390,8 +390,14 @@ func load[T client.Object](t T, path, name string) (T, error) {
 func assureResource[T client.Object](ctx context.Context, r *BpfmanConfigReconciler,
 	bpfmanConfig *v1alpha1.Config, resource T, needsUpdate func(existing T, desired T) bool) error {
 	if isOverridden(resource, bpfmanConfig.Spec.Overrides, r.Scheme) {
+		// Resolve GVK via the scheme to get the kind reliably
+		kind := ""
+		gvks, _, err := r.Scheme.ObjectKinds(resource)
+		if err == nil && len(gvks) > 0 {
+			kind = gvks[0].Kind
+		}
 		r.Logger.Info("Skipping unmanaged resource (override in place)",
-			"kind", resource.GetObjectKind().GroupVersionKind().Kind,
+			"kind", kind,
 			"namespace", resource.GetNamespace(), "name", resource.GetName())
 		return nil
 	}


apis/v1alpha1/config_types.go
```diff
@@ -57,6 +57,17 @@ type ConfigSpec struct {
 	// Namespace holds the namespace where bpfman-operator resources shall be
 	// deployed.
 	Namespace string `json:"namespace,omitempty"`
+
+	// overrides is list of overides for components that are managed by
```
frobware: typo: s/overides/overrides

controllers/bpfman-operator/config_test.go
```diff
@@ -578,3 +614,136 @@ func verifyCM(cm *corev1.ConfigMap, requiredFields map[string]*string) error {
...
+		return err
+	}
+
+	bpfmanConfig.Spec.Overrides = []v1alpha1.ComponentOverride{
+		{
+			Kind:      "ConfigMap",
+			Group:     "",
+			Namespace: "bpfman",
+			Name:      "bpfman-config",
+			Unmanaged: true,
```
frobware: Post #461, isn't the configmap now an implementation detail? Should we allow for the configmap to be unmanaged?

frobware (changes requested):
ConfigMap should not be overrideable given that it's now an internal implementation detail of the Config CRD (post #461). Should `isOverrideable` narrow the list down to the DaemonSet, CSIDriver, SCC -- everthing else is excluded.

PR #477, opened 2025-09-22
Fix overlaying bpf mount

nocturo:
The check in the initContainer for the deployed `daemonSet` is only looking for exact text from the mount command, which leads to false positives and `bpfman-operator` mounting an overlay bpf filesystem causing problems with other applications relying on it. (I had issues with Cilium but it will break anything that is using it)

Below is an excerpt of what I mean, first we check the initial state using current method:
```bash
root@test:/# /bin/mount | /bin/grep 'bpffs on /sys/fs/bpf'
root@test:/# echo $?
1
root@test:/# ls /sys/fs/bpf/
cilium  tc
root@test:/# mount|grep bpf
bpf on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
root@test:/#
```

then after `bpfman-operator` has been deployed:
```bash
root@test:/# mount|grep bpf
bpf on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
bpffs on /sys/fs/bpf type bpf (rw,relatime)
```

This PR addresses the problem by using `findmnt` for finding the right mount type for `/sys/fs/bpf` mountpoint. After changing my `daemonSet` to this, I no longer have issues running `bpfman-operator` and Cilium.


frobware:
@nocturo - Just a quick heads-up and courtesy call: I’ve raised https://github.com/bpfman/bpfman-operator/pull/490. I wanted to mention this since you previously raised and fixed the bpffs mount issue. The goal of #490 is to create a Go implementation that performs the equivalent functions of findmnt/mount. This means we can eliminate the need for an init container that requires a Fedora image, as the functionality will now be integrated into the agent. From your perspective, nothing should change...

PR #516, opened 2026-05-05
Add setup-envtest to tools.go for vendoring

alebedev87:
Add `tools/tools.go` with a blank import of `setup-envtest` so the tool is included in the vendor directory. Uses the `release-0.22` branch pseudo-version (`v0.0.0-20260125163108-a19ec76a3c5d`) which requires Go 1.24 (compatible with the project's Go 1.25). The only tagged semver release (`v0.24.0`) requires Go 1.26 and is not usable yet.

Replace the `go install`/binary invocation of `setup-envtest` in the Makefile with `go run`, removing the `envtest` target and `ENVTEST` variable. Since the tool is now vendored, `go run` picks it up directly without needing a separate install step.

Also includes the vendor drift CI check from #517: run `go mod tidy && go mod vendor` and assert with `git status --porcelain` that no tracked modifications or untracked entries remain under `vendor/`, `go.mod` or `go.sum`.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -177,7 +177,15 @@ jobs:
         run: make verify
 
       - name: Check clean vendors
-        run: go mod vendor
+        run: |
+          go mod tidy
+          go mod vendor
+          if [ -n "$(git status --porcelain vendor go.mod go.sum)" ]; then
+            echo "vendor/, go.mod or go.sum is out of sync with 'go mod tidy && go mod vendor':"
+            git status --porcelain vendor go.mod go.sum
+            git diff -- vendor go.mod go.sum
+            exit 1
+          fi
 
       - name: Verify generated bundle manifest
         run: |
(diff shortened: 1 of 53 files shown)


frobware:
Two observations from comparing against #518.

**1. Makefile `$(shell ...)` still masks `setup-envtest` failures**

```make
KUBEBUILDER_ASSETS="$(shell go run sigs.k8s.io/controller-runtime/tools/setup-envtest use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
```

`$(shell ...)` captures stdout but discards the exit code, so if `go run` fails (e.g. `cannot find module providing package ... import lookup disabled by -mod=vendor`) the variable expands to `""` and `go test` proceeds. No test currently consumes `KUBEBUILDER_ASSETS`, so it passes silently. The vendor-cleanliness check from `66171cec` would catch this specific drift but not other setup-envtest failures that leave vendor clean. Moving the substitution into the recipe shell and asserting `KUBEBUILDER_ASSETS` is non-empty closes that gap -- see `04c7abc7` on #518.

**2. Commit message for `66171cec` could note why `git diff --exit-code` is not sufficient**

It ignores untracked files -- which is exactly the failure mode here (newly materialised vendor dirs that were never `git add`ed). Worth a line in the body so future `git blame` answers itself.

alebedev87:
Force pushed to retest the bundle deployment on KIND.

frobware (approved):
LGTM
Thanks!

frobware (approved):
LGTM
Thanks!

frobware (approved):
LGTM
Thanks!

PR #521, opened 2026-05-07
Move gen tools to `tools/tools.go` and use `go run`

alebedev87:
Consolidate tool dependency tracking from `hack/tools.go` and Makefile `go install @version` targets into `tools/tools.go` using the blank import pattern. This pins tool versions in `go.mod` and vendors them, matching the pattern already used for `setup-envtest`.

Tools moved: `controller-gen`, `register-gen`, `informer-gen`, `lister-gen`, `client-gen`. Makefile now invokes them via `go run` instead of installing to `bin/`. Removed unused `CM_VERIFIER` variable, `CONTROLLER_TOOLS_VERSION`, and `K8S_CODEGEN_VERSION`.


frobware (approved):
LGTM
Thanks!

PR #523, opened 2026-05-11
Bump kustomize from v3.8.7 to v5.8.1 and add to `tools/tools.go`

alebedev87:
Move kustomize to `tools/tools.go` and use `go run` for invocation, consistent with how `controller-gen`, `client-gen`, and other tools are already managed. This replaces the curl-based install script that downloaded a prebuilt binary into `bin/`.

The `KUSTOMIZE_VERSION`, `KUSTOMIZE_INSTALL_SCRIPT`, and the `kustomize` Makefile target are removed. All targets that depended on the `kustomize` prerequisite are updated since `go run` handles tool resolution automatically.


frobware (approved):
LGTM

PR #532, opened 2026-06-03
bpfman-operator: add node placement fields to Config daemon spec

### LINKED ISSUE #531: Config: allow tolerations for the bpfman daemon DaemonSets
### What would you like to be added?

An optional `tolerations` field on the `Config` CRD, so the bpfman daemon and metrics-proxy DaemonSets can schedule onto nodes with custom taints without resorting to the `spec.overrides` escape hatch.

```yaml
kind: Config
metadata:
  name: bpfman-config
spec:
  namespace: bpfman
  tolerations:
    - key: workload
      operator: Equal
      value: gpu
      effect: NoSchedule
```

### Why is this needed?

The daemon DaemonSet is rendered from a static manifest (`config/bpfman-deployment/daemonset.yaml`) whose tolerations are hardcoded to the control-plane/master taints. On clusters with custom taints — infrastructure pools, GPU pools, dedicated workloads — the daemon cannot reach those nodes at all.

That is not a partial degradation: the operator expects a `BpfApplicationState` object from every node, so a single uncovered node leaves **every** `BpfApplication` in the cluster stuck in `Pending`.

This was previously achievable when the daemon was deployed via a ConfigMap containing the raw DaemonSet YAML. With the move to the `Config` CRD the only remaining option is marking the DaemonSet `unmanaged` via `spec.overrides`, which disables all reconciliation for it.

### Proposed behavior

- `tolerations` lives on `ConfigSpec` rather than `DaemonSpec`, since it must reach the metrics-proxy DaemonSet too.
- Configured tolerations are **merged** into both DaemonSets on top of the static manifest's, skipping duplicates. Replacing would drop the built-in control-plane/master tolerations and hit the `Pending` problem described above.

### Note

This issue originally also covered `nodeSelector` and `affinity`. Restricting the daemon to a subset of nodes hits the same node-coverage invariant and needs a dedicated effort — that is now tracked in #542.

huseyinbabal:
## Summary

Add optional `nodeSelector`, `tolerations`, and `affinity` fields to the `Config` CRD under `spec.daemon`, so users can control where the bpfman daemon pods are scheduled.

Fixes #531

## Motivation

The daemon DaemonSet is rendered from a static manifest whose tolerations are hardcoded to the control-plane/master taints. There was no first-class way to target the daemon at specific nodes, or to allow it onto nodes with custom taints. This was possible when the daemon was deployed via a raw ConfigMap, but that ability was lost in the move to the `Config` CRD (the only remaining option was the heavy `spec.overrides` unmanaged escape hatch).

## Changes

- Added `nodeSelector`, `tolerations`, and `affinity` to `DaemonSpec`.
- The operator applies them to both the bpfman daemon and metrics-proxy
  DaemonSets, with per-field replace semantics:
  - omitted → the static manifest default is preserved (e.g. the built-in
    control-plane/master tolerations stay),
  - set (including an empty `tolerations: []`) → replaces the static value.
- Regenerated CRD, deepcopy, and bundle manifests; updated the sample.
- Added unit tests for both DaemonSets and the defaults-preserved case.

## Example

```yaml
kind: Config
metadata:
  name: bpfman-config
spec:
  daemon:
    image: quay.io/bpfman/bpfman:latest
    nodeSelector:
      kubernetes.io/os: linux
    tolerations:
      - operator: Exists
    affinity:                       # optional: keep the daemon off control-plane nodes
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: node-role.kubernetes.io/control-plane
                  operator: DoesNotExist
```

## Test plan

- `make test` and `make verify` pass.
- Verified on a multi-node kind cluster: with `tolerations: [{operator: Exists}]` the daemon schedules onto the tainted control-plane node; removing it keeps the daemon off the control-plane.


huseyinbabal:
It seems, integration tests action is failing for a while already

frobware:
> It seems, integration tests action is failing for a while already

Fixed in https://github.com/bpfman/bpfman-operator/pull/537.

alebedev87:
Thanks for the contribution, @huseyinbabal! The hardcoded tolerations in the static manifest have been a friction point and we appreciate you addressing it.

Before we proceed I have a few questions/points to discuss:

- `nodeSelector` and `affinity`: could you share the concrete scenarios that motivated including these? We're looking for strong use cases, especially for `nodeSelector`. For affinity, I'm hesitant - it brings a large amount of generated CRD schema (~900 lines) and some of its fields overlap with `nodeSelector`, which as noted below is something we'd like to address separately and more carefully.
- `tolerations` are confirmed needed: we hit this problem in our own internal testing - nodes with custom taints (infrastructure pools, GPU pools, dedicated workloads) can't get the daemon today without reaching for `spec.overrides`.
- Why `nodeSelector` is harder than it looks: the operator contains an invariant - it lists all cluster nodes and expects a `BpfApplicationState` object from each one, which is how it determines that reconciliation is complete across the cluster. If the daemon is restricted to a subset of nodes via `nodeSelector`, nodes without an agent never produce their state objects and any `BpfApplication` ends up stuck in `Pending` with no actionable error.  Solving this properly requires a more significant rework of how the operator tracks daemon node coverage. I'd like this to be approached as a dedicated issue.

Would you be willing to repurpose this PR to add `tolerations` only? Also rather than putting them in `config.spec.daemon.tolerations`, I think `config.spec.tolerations` is cleaner. `DaemonSpec` fields currently apply exclusively to the bpfman-daemon `DaemonSet`, but `tolerations` need to reach the metrics-proxy `DaemonSet` too. Placing `tolerations` at the `ConfigSpec` level reflects that correctly, and also gives us the natural extension point for `nodeSelector` in the future.

huseyinbabal:
> Thanks for the contribution, @huseyinbabal! The hardcoded tolerations in the static manifest have been a friction point and we appreciate you addressing it.
> 
> Before we proceed I have a few questions/points to discuss:
> 
> * `nodeSelector` and `affinity`: could you share the concrete scenarios that motivated including these? We're looking for strong use cases, especially for `nodeSelector`. For affinity, I'm hesitant - it brings a large amount of generated CRD schema (~900 lines) and some of its fields overlap with `nodeSelector`, which as noted below is something we'd like to address separately and more carefully.
> * `tolerations` are confirmed needed: we hit this problem in our own internal testing - nodes with custom taints (infrastructure pools, GPU pools, dedicated workloads) can't get the daemon today without reaching for `spec.overrides`.
> * Why `nodeSelector` is harder than it looks: the operator contains an invariant - it lists all cluster nodes and expects a `BpfApplicationState` object from each one, which is how it determines that reconciliation is complete across the cluster. If the daemon is restricted to a subset of nodes via `nodeSelector`, nodes without an agent never produce their state objects and any `BpfApplication` ends up stuck in `Pending` with no actionable error.  Solving this properly requires a more significant rework of how the operator tracks daemon node coverage. I'd like this to be approached as a dedicated issue.
> 
> Would you be willing to repurpose this PR to add `tolerations` only? Also rather than putting them in `config.spec.daemon.tolerations`, I think `config.spec.tolerations` is cleaner. `DaemonSpec` fields currently apply exclusively to the bpfman-daemon `DaemonSet`, but `tolerations` need to reach the metrics-proxy `DaemonSet` too. Placing `tolerations` at the `ConfigSpec` level reflects that correctly, and also gives us the natural extension point for `nodeSelector` in the future.

Thanks @alebedev87 — agreed on all points. Happy to repurpose this to
`tolerations` only under `config.spec.tolerations`.

Our `nodeSelector` scenario, for the record: our workers are split into a
traffic-serving pool that uses bpfman, and an internal pool (build agents, CI,
internal tooling) that never needs eBPF. We already scope programs with
`BpfApplication.spec.nodeSelector`, so on the internal pool the agent loads
nothing, it just idles with a `NotSelected` state. But those nodes still run 2
DaemonSet pods with 4 privileged containers, `hostNetwork: true`, and
bidirectional mounts of `/sys/fs/bpf` and `/var/lib/kubelet/pods`. The resource
cost is secondary; the real driver is that we don't want privileged host-network
workloads on nodes with no functional need for them.

As you say, exposing `nodeSelector` alone wouldn't fix that, it'd just wedge
`BpfApplication`s in `Pending`. The actual ask is partial daemon node coverage,
i.e. deriving the expected node set from where the daemon is scheduled rather
than an unconditional `List` of all nodes. Happy to open a dedicated issue for it.

One question on tolerations: merge or replace? The static manifests hardcode the
control-plane/master tolerations, so with replace semantics adding a GPU taint
toleration would silently drop them and pull the daemon off control-plane nodes.

alebedev87:
> Our nodeSelector scenario, for the record: our workers are split into a
traffic-serving pool that uses bpfman, and an internal pool (build agents, CI,
internal tooling) that never needs eBPF. We already scope programs with
BpfApplication.spec.nodeSelector, so on the internal pool the agent loads
nothing, it just idles with a NotSelected state. But those nodes still run 2
DaemonSet pods with 4 privileged containers, hostNetwork: true, and
bidirectional mounts of /sys/fs/bpf and /var/lib/kubelet/pods. The resource
cost is secondary; the real driver is that we don't want privileged host-network
workloads on nodes with no functional need for them.
 
The use case seems reasonable, I think it makes sense to support it. Let's track it as a dedicated issue so it doesn't block the toleration work, which can go in independently (and faster too).

> One question on tolerations: merge or replace? The static manifests hardcode the
control-plane/master tolerations, so with replace semantics adding a GPU taint
toleration would silently drop them and pull the daemon off control-plane nodes.

Good question. Replace semantics is arguably cleaner, but removing the built-in control-plane/master tolerations would have the same effect as nodeSelector restriction - the daemon stops running on those nodes, and `BpfApplication` ends up stuck in `Pending` state. Fixing that invariant can be the effort included in the support of `nodeSelector`.  Merge semantics should be enough to answer the needs without breaking the invariant. We can tighten this to replace semantics later - the API is still alpha.

huseyinbabal:
Thanks @alebedev87 — repurposed as discussed.

- `tolerations` only, moved to `config.spec.tolerations` so it also reaches the
  metrics-proxy DaemonSet.
- Merge semantics: the configured tolerations are appended to the ones in the
  static manifests, and tolerations already present are skipped, so re-declaring
  the control-plane/master ones is a no-op.
- `nodeSelector` and `affinity` dropped — the generated CRD schema is down from
  ~970 lines to 44.

I opened #542 for the node-coverage rework and the `nodeSelector` support, and
narrowed #531 down to tolerations so this PR closes it.

PR #536, opened 2026-06-22
Prefer `KUBECONFIG` over in-cluster config in `GetK8sConfigOrDie`

alebedev87:
When running inside a cluster with `KUBECONFIG` env var set (e.g. integration tests in a Tekton pod against a remote cluster), prefer the kubeconfig over the in-cluster service account config. The default in-cluster behavior is unchanged when `KUBECONFIG` is not set.

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -18,6 +18,7 @@ package helpers
 
 import (
 	"fmt"
+	"os"
 
 	bpfmaniov1alpha1 "github.com/bpfman/bpfman-operator/apis/v1alpha1"
 	bpfmanclientset "github.com/bpfman/bpfman-operator/pkg/client/clientset"
@@ -101,8 +102,9 @@ func (t TcProgramDirection) String() string {
 
 var log = ctrl.Log.WithName("bpfman-helpers")
 
-// getk8sConfig gets a kubernetes config automatically detecting if it should
-// be the in or out of cluster config. If this step fails panic.
+// GetK8sConfigOrDie gets a kubernetes config automatically detecting if it should
+// be the in or out of cluster config. When running in-cluster, KUBECONFIG
+// env var takes precedence over the in-cluster config. If this step fails panic.
 func GetK8sConfigOrDie() *rest.Config {
 	config, err := rest.InClusterConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/helpers/helpers.go
```diff
@@ -101,8 +102,9 @@ func (t TcProgramDirection) String() string {
 
 var log = ctrl.Log.WithName("bpfman-helpers")
 
-// getk8sConfig gets a kubernetes config automatically detecting if it should
-// be the in or out of cluster config. If this step fails panic.
+// GetK8sConfigOrDie gets a kubernetes config automatically detecting if it should
+// be the in or out of cluster config. When running in-cluster, KUBECONFIG
+// env var takes precedence over the in-cluster config. If this step fails panic.
```
frobware, line 107: Reading the doc comments I read this as "KUBECONFIG wins" so why don't we honour that path first? 

```go
func GetK8sConfigOrDie() *rest.Config {
	// An explicit KUBECONFIG overrides the in-cluster config; without it we
	// prefer in-cluster and fall back to the default kubeconfig out of cluster.
	if os.Getenv("KUBECONFIG") == "" {
		// A non-nil error here means we are not in a cluster
		// (rest.ErrNotInCluster); fall through to the kubeconfig below.
		if config, err := rest.InClusterConfig(); err == nil {
			log.Info("Program running inside the cluster, picking the in-cluster configuration")
			return config
		}
	}

	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{}).ClientConfig()
	if err != nil {
		panic(err)
	}

	log.Info("Program picking config from kubeconfig", "host", config.Host)
	return config
}
```

untested - just an observation from reading the function commentary.

alebedev87: > Reading the doc comments I read this as "KUBECONFIG wins"
```
When running in-cluster, KUBECONFIG env var takes precedence over the in-cluster config.
```
The comment says that `KUBECONFIG` wins in the second case when we are "in cluster". The other case is when we are "out of cluster".

> so why don't we honour that path first? untested - just an observation from reading the function commentary.

Yes, I thought of this approach too. I wanted to remain as less intrusive as possible to not break anything/anyone. Plus `GetDefaultFilename()` does a little more than just reading `KUBECONFIG` variable, I don't need to break this flow for "out of cluster" case.

frobware (approved):
LGTM
Thanks!

